PR #1, opened 2018-07-12
[WIP] Initial Work on Layout and Start Buttons

wgranger:
This PR began with some work from pairing exercises while Florian was visiting. Many of the draggable and interactive element classes/interfaces are leftover from that work and will likely be used as the app progresses, so I left it in the code.

I stopped this code at the point where it had toggle functionality of buttons. There is still a good bit of work that needs to be done with functionality and styling, but I decided it was best to make a commit now before the code changes got too large with the initial commit.

For future PRs, I do plan to work on a forked branch. This initial commit is straight to the Zooniverse branch.

WIP - as some clean up would help before a merge.


noahmalmed:
Not sure if this is on purpose, but there is a .sln file at the root level of the project and one under `./GalaxyZooTouchTable`. The root one doesn't seem to load correctly in VS

wgranger:
Ahhh. No, that was totally a mistake on my end, thanks for catching it.

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/TableUserFactory.cs
```diff
@@ -0,0 +1,82 @@
...
+        Heart
+    }
+
+    abstract class TableUser
+    {
+        public abstract string ThemeColor { get; }
+        public abstract BitmapImage StartButton { get; }
+    }
+
+    class PersonUser : TableUser
```
noahmalmed: As the code is right now, someone consuming this could instantiate their own `PersonUser` or `StarUser` etc without using the factory method. I think for all these subclasses you may want to add a private constructor like `private PersonUser() {}`. That way consumers of these models have to use the factory method.

GalaxyZooTouchTable/GalaxyZooTouchTable/MainWindow.xaml.cs
```diff
@@ -0,0 +1,40 @@
...
+        private void SetUserContext()
+        {        
+           TableUser personUser = TableUserFactory.Create(UserType.Person);
+           TableUser lightUser = TableUserFactory.Create(UserType.Light);
+           TableUser starUser = TableUserFactory.Create(UserType.Star);
+           TableUser heartUser = TableUserFactory.Create(UserType.Heart);
+           TableUser faceUser = TableUserFactory.Create(UserType.Face);
+           TableUser earthUser = TableUserFactory.Create(UserType.Earth);
+
+           PersonUser.DataContext = personUser;
```
noahmalmed: For my sake, could you explain what is going here?

GalaxyZooTouchTable/GalaxyZooTouchTable/UserConsole.xaml.cs
```diff
@@ -0,0 +1,28 @@
...
+    /// <summary>
+    /// Interaction logic for UserConsole.xaml
+    /// </summary>
+    public partial class UserConsole : UserControl
+    {
+        ClassificationPanel Classifier { get; set; }
+
+        public UserConsole()
+        {
+            InitializeComponent();
```
noahmalmed: Is this WIP stuff? Or am I missing where it's defined

GalaxyZooTouchTable/GalaxyZooTouchTable/MainWindow.xaml.cs
wgranger: The app starts with several components already initialized (`UserConsole`). Once the app opens, the `MainWindow` component is responsible for creating several user instances (via the factory) and attaching them to the already rendered `UserConsoles`. Adding the TableUsers to the DataContext of each component allows their properties to be accessible through data binding.

GalaxyZooTouchTable/GalaxyZooTouchTable/UserConsole.xaml.cs
wgranger: This will change a bit, but the code here is pretty simple for now. By default, a xaml item only initializes itself through the constructor when added to a project in WPF.

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/TableUserFactory.cs
wgranger: Yes. This makes total sense. Good eye.

wgranger:
Has been updated per your request. Let me know if there are any other outstanding issues with this branch.

PR #2, opened 2018-07-18
Style Classification Panel

wgranger:
This PR makes a few changes to the layout and also styles the classifier panel to look like the prototype. Future updates will include adding standalone functional components to the classifier.


GalaxyZooTouchTable/GalaxyZooTouchTable/ClassificationPanel.xaml.cs
```diff
@@ -12,14 +16,29 @@ public partial class ClassificationPanel : UserControl
...
+        {
+            float StartPos = Console.ClassifierOpen ? -600 : 0;
+            float EndPos = Console.ClassifierOpen ? 400 : -600;
+
+            var PanelTransform = new TranslateTransform(0, StartPos);
+            RenderTransform = PanelTransform;
+            DoubleAnimation panelAnimation = new DoubleAnimation(EndPos, new Duration(TimeSpan.FromSeconds(0.25)));
+            if (Console.ClassifierOpen)
+            {
+                panelAnimation.Completed += (s, e) => Console.ControlPanel.Children.Remove(this);
```
noahmalmed: I know your not using them, but could you use more descriptive names for `s` and `e`

GalaxyZooTouchTable/GalaxyZooTouchTable/UserConsole.xaml.cs
```diff
@@ -10,19 +14,45 @@ namespace GalaxyZooTouchTable
     public partial class UserConsole : UserControl
     {
         ClassificationPanel Classifier { get; set; }
+        public bool ClassifierOpen { get => classifierOpen; set => classifierOpen = value; }
```
noahmalmed: Couple of things here:
1) Why do you have a public and private version of this property? Seems like it would just be fine if you left this as a `public bool` and got rid of the `private` one.

2) Not a huge deal, but I think the name of this property is kind of confusing. `ClassifierOpen` sounds like an action rather than a property. It may be more clear to name this something like `IsClassifierOpen` or `ClassifierIsOpen`.

GalaxyZooTouchTable/GalaxyZooTouchTable/UserConsole.xaml.cs
```diff
@@ -10,19 +14,45 @@ namespace GalaxyZooTouchTable
...
 
         private void StartButton_TouchUp(object sender, TouchEventArgs e)
+        {
+            ToggleClassifier();
+        }
+
+        public void ToggleClassifier()
+        {
+            MoveButton();
+            AddClassifier();
```
noahmalmed: Is this right? You add a classifier every time you toggle the classifier?

GalaxyZooTouchTable/GalaxyZooTouchTable/ClassificationPanel.xaml.cs
```diff
@@ -12,14 +16,30 @@ public partial class ClassificationPanel : UserControl
...
+            Console.ClassifierOpen = !Console.ClassifierOpen;
+        }
+
+        public void MoveClassifier()
+        {
+            float StartPos = Console.ClassifierOpen ? -300 : 0;
+            float EndPos = Console.ClassifierOpen ? 200 : -300;
+
+            var PanelTransform = new TranslateTransform(0, StartPos);
+            RenderTransform = PanelTransform;
```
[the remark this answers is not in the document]
wgranger: RenderTransform is a way or transforming components in WPF. In this instance, the `this` keyword is implied, so `this.RenderTransform` is the effect this is having.

GalaxyZooTouchTable/GalaxyZooTouchTable/UserConsole.xaml.cs
wgranger: Yes. My thought here is that a classifier will be removed when a session ends and a new classifier will introduce a new session. However, this could change as I start to add functionality to the classifier.

wgranger:
Alright. Several changes were made to this PR. Basically, these components should be more responsive in general, and that means choosing UI elements and panels that are in line with WPF layout standards. There are many line changes in the most recent commits that remove absolute styling to keep the app more flexible when testing the app on small and large screens alike. However, the app should perform in a 16:9 aspect ratio layout (same as our touch table).

GalaxyZooTouchTable/GalaxyZooTouchTable/UserConsole.xaml.cs
noahmalmed: Are you going to make these changes?

wgranger: Yes. Will comment below.

wgranger:
Still need to make a change here with the private getter/setter in `UserConsole`. There is also a bug currently where the app doesn't function on multiple systems due to a broken reference path from the `FontAwesome` reference, so a fix to that will also come in a forthcoming commit.

wgranger:
Alright. Changes were made per your request. I also added a setup project that handles publishing the project. The bug I experienced today involved using the `FontAwesome` resource .DLL. Basically, when trying to open the bin/debug .exe file for the project, the resource wasn't being included. In order to open the app on another system, the project should be published, which is why the setup folder is included. I'll push the current setup app to the Google drive now.

PR #5, opened 2018-08-10
Style Classifier with Correct Sizing

wgranger:
With the previous PR, there were still a few issues remaining concerning sizing of the classifier on small screens vs. the large touchtable. This PR takes care of those issues and also adds the `Centerpiece` component messaging to users.

The setup should also be working, and I can walk you through that if needed.


noahmalmed:
image

Looks like there are some sizing issues with the middle image cutting off some of the other views. Seems like you can just fix it by making the background of the center image transparent.

noahmalmed:
Also looks like the bar that goes about "Help Classify the Galaxy" is going over the avatar circle. Is that correct?

Additionally it looks like the avatar circle is overlapping with the "Close workspace button"

wgranger:
Alright. Changes made for your request. This should handle the avatar sizing correctly for multiple windows.

noahmalmed:
Looks good! Thanks

PR #6, opened 2018-08-29
Submit Classifications

wgranger:
This PR allows subjects to be loaded and classifications sent via the app. 

**Note**: some changes were needed to the client. I've pushed those changes up to the Client repo, but they should also be reflected here via a local dependency. Once the changes are merged in the client repo I'll update the Nuget package and relink the client package here to that updated version.


GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ClassificationPanelViewModel.cs
```diff
@@ -0,0 +1,174 @@
...
+        public event PropertyChangedEventHandler PropertyChanged;
+
+        private void OnPropertyRaised(string propertyname)
+        {
+            PropertyChanged(this, new PropertyChangedEventArgs(propertyname));
+        }
+
+        private void LoadCommands()
+        {
+            SelectAnswer = new CustomCommand(ChooseAnswer, CanSelectAnswer);
```
noahmalmed: Can you explain how `CustomCommand` works?

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ClassificationPanelViewModel.cs
```diff
@@ -0,0 +1,174 @@
...
+            SubmitClassification = new CustomCommand(SendClassification, CanSendClassification);
+        }
+
+        private async void SendClassification(object obj)
```
noahmalmed: For all of the following methods, can you use a more descriptive variable name than `obj` it's kind of unclear what that variable is for

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ClassificationPanelViewModel.cs
```diff
@@ -0,0 +1,174 @@
...
+        private bool CanSelectAnswer(object obj)
+        {
+            return CurrentAnswers.Count > 0;
+        }
+
+        public List<AnswerButton> ParseTaskAnswers(List<TaskAnswer> answers)
+        {
+            List<AnswerButton> renderedAnswers = new List<AnswerButton>();
+
+            for (int i = 0; i < answers.Count; i++)
```
noahmalmed: Can you use a more descriptive name than `i` here?

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ClassificationPanelViewModel.cs
```diff
@@ -0,0 +1,174 @@
...
+            CurrentClassification.Links.Subjects.Add(subject.Id);
+
+            CurrentAnnotation = null;
+            SelectedItem = null;
+            CommandManager.InvalidateRequerySuggested();
+        }
+
+        private async void GetSubject()
+        {
+            if (Subjects.Count > 0)
```
noahmalmed: Not a huge deal, but you can change this to
```

if(Subjects.Count <= 0) 
{
	ApiClient client = new ApiClient();
    NameValueCollection query = new NameValueCollection
	{
		{ "workflow_id", Config.WorkflowId }
	};
	Subjects = await client.Subjects.GetList("queued", query);
}

SetSubject()
```

Also, you could then just get ride of `setSubject` and just put that code inline

GalaxyZooTouchTable/GalaxyZooTouchTable/Lib/Utilities.cs
```diff
@@ -0,0 +1,21 @@
+﻿using Newtonsoft.Json.Linq;
+using PanoptesNetClient.Models;
+using System.Linq;
+
+namespace GalaxyZooTouchTable.Lib
+{
+    public static class Utilities
+    {
+        public static string GetSubjectLocation(Subject subject, int frame = 0)
```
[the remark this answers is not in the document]
wgranger: Yes, I think that change should work. I'll have to update the client instead of changing code here.

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/AnswerButton.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        public string Label { get; set; }
+        public string Url { get; set; }
+        public int Index { get; set; }
+
+        public AnswerButton(TaskAnswer answer, int index)
+        {
+            var UrlMatch = Regex.Match(answer.Label, @"((http|ftp|https):\/\/[\w\-_]+(\.[\w\-_]+)+([\w\-\.,@?^=%&amp;:/~\+#]*[\w\-\@?^=%&amp;/~\+#])?)");
+            string removedBrackets = Regex.Replace(answer.Label, @"!\[(.*?)\]", "");
+            string removedUrl = Regex.Replace(removedBrackets, @"\(" + UrlMatch + @"(.*?)\)", "");
```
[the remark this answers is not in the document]
wgranger: Yeah, so in the Galaxy Zoo workflow the workflow answers contain markdown for an image, and there aren't any good/simple ways to transform markdown to XAML, so this handles the regex to find the image link in the workflow answers. I can put a comment here and change the variable names to make things more apparent.

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/TableUserFactory.cs
```diff
@@ -13,11 +13,11 @@ enum UserType
         Heart
     }
 
-    abstract class TableUser
+    public class TableUser
```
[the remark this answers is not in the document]
wgranger: This was changed because I'm thinking of using an `ObservableCollection<TableUser>` group in a future PR to track when users come to the table. This will help the main app level know who is at the table for purposes of asking X user for help with a classification and knowing what to display on the center of the app. 

However, using `ObservableCollection<TableUser>` with an abstract is not allowed due to accessibility levels of an abstract class.

GalaxyZooTouchTable/GalaxyZooTouchTable/NullImageConverter.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using System.Globalization;
+using System.Windows;
+using System.Windows.Data;
+
+namespace GalaxyZooTouchTable
+{
+    public class NullImageConverter : IValueConverter
```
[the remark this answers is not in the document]
wgranger: This class allows a null image to be "rendered" if the subject hasn't yet been fetched from the API. WPF freaks out when binding images if the source isn't yet available.

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ClassificationPanelViewModel.cs
wgranger: `CustomCommand` handles a lot of the necessary actions required by a command. Usually, with an MVVM setup in WPF, functions are triggered by a command, and it's common for commands to run through a custom class derived from `ICommand`. This provides two functions: executing the main functionality of the command (`ChooseAnswer` above) and determining if a command should be disabled (`CanSelectAnswer` above). 

There are a number of ways to do this, and the `ICommand` interface expects these two `ChooseAnswer` and `CanSelectAnswer` functions. You can find more info https://www.codeproject.com/Tips/813345/Basic-MVVM-and-ICommand-Usage-Example

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ClassificationPanelViewModel.cs
wgranger: Of course. The `object obj` param is necessary to satisfy the `ICommand` interface, but changing it to `object sender` would help.

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ClassificationPanelViewModel.cs
wgranger: I can. `i` is often used as shorthand for `index`, but it can easily be changed to `index`.

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/TableUserFactory.cs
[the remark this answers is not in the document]
noahmalmed: Would it work if TableUser was an interface or was private? In order to enforce using the factory class you made, I think it's important that you can't just instantiate a table user without any of the predefined values

GalaxyZooTouchTable/GalaxyZooTouchTable/NullImageConverter.cs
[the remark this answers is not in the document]
noahmalmed: Could you add a little comment explaining that? It's not really immediately clear what the class does based on the code

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/TableUserFactory.cs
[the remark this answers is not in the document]
wgranger: I'll try one of those options. I definitely agree that is a better option.

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ClassificationPanelViewModel.cs
noahmalmed: cool that makes sense

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ClassificationPanelViewModel.cs
noahmalmed: Haha, sorry I'm so ticky tacky about names. You can leave it as `i`, I just think the code becomes a little more readable when variable names are more verbose. Even something like `answerIndex` is so much more readable to me than `i`

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ClassificationPanelViewModel.cs
```diff
@@ -102,19 +102,19 @@ private async void SendClassification(object obj)
...
+        private bool CanSendClassification(object sender)
         {
             return CurrentAnnotation != null;
         }
 
-        private void ChooseAnswer(object obj)
+        private void ChooseAnswer(object sender)
         {
-            AnswerButton button = (AnswerButton)obj;
+            AnswerButton button = (AnswerButton)sender;
```
noahmalmed: Should there be something here to handle if the sender isn't castable to `AnswerButton`?

wgranger: I should be able to change this to `AnswerButton button= sender as AnswerButton;` to return `null` if uncastable rather than an exception.

noahmalmed:
The code looks good. I'm going to test it tomorrow morning when I'm in the office

noahmalmed:
Everything running correctly. Merging 👍

PR #8, opened 2018-09-17
Close Workspace Dialog

wgranger:
This PR introduces the close dialog into the touch table.

Note: there are several items I would like to change here (namely the view models reliance on other UI elements) to adhere to MVVM patterns. However, this would cause a good bit of changes in the structure of the code. For now, I'm aiming towards functionality, and then when we reach phase one I plan to go back over these elements to give a facelife and adhere more closely to MVVM principles. 

Also, the QR code is not yet visible as that asset is not available on Atomic or InVision.


noahmalmed:
Code looks good, everything looked good testing as well

PR #20, opened 2018-10-23
Update Examples Panel

### LINKED ISSUE #15: Refactor Examples Pullout
The examples pullout was originally written without strict adherence to MVVM. There is also some odd animation logic in the code-behind. This should be cleaned up, along with making the hit points more user friendly.

Also, the images should enlarge when a category is selected.

wgranger:
This PR refactors the examples panel, which was previously in a pretty sad state. 

Although there are more lines here, the confusing code has been removed, and the heavy lifting of the animations logic is covered in the XAML.

Close #15


GalaxyZooTouchTable/GalaxyZooTouchTable/Views/ExamplesPanel.xaml
```diff
@@ -133,128 +316,238 @@
...
-                <Grid Grid.Row="1" Grid.Column="1" Panel.ZIndex="1" Style="{StaticResource SelectedItemGrid}">
+                <Grid x:Name="ExampleGrid" Grid.Column="1" Grid.Row="2">
                     <Grid.RowDefinitions>
-                        <RowDefinition Height="1*"/>
+                        <RowDefinition Style="{StaticResource ActiveExample}" Height="1*"/>
                         <RowDefinition Height="2*"/>
                     </Grid.RowDefinitions>
 
-                    <Border Grid.Row="0" Margin="2" x:Name="SelectedElement"/>
+                    <Border Style="{StaticResource SmoothExampleBorder}" BorderThickness="0 1 0 0" BorderBrush="{StaticResource DarkGrayColor}">
```
wgranger: These three `Border`s (`SmoothExampleBorder`, `FeaturesExampleBorder`, and `StarExampleBorder`) are the one place where I have some repetitive code I could _maybe_ refactor into a template. Note, the borders are the same with small differences in the bindings and Y render transform locations.

I'll continue to play with this a bit, but the rest is good for a complete review.

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/GalaxyExample.cs
```diff
@@ -8,47 +7,64 @@ public enum GalaxyType
...
-                    break;
-                case GalaxyType.Star:
-                    ImageOne = new BitmapImage(new Uri("../Images/GalaxyExamples/star1.jpg", UriKind.Relative));
-                    ImageTwo = new BitmapImage(new Uri("../Images/GalaxyExamples/star2.jpg", UriKind.Relative));
-                    ImageThree = new BitmapImage(new Uri("../Images/GalaxyExamples/smooth3.jpg", UriKind.Relative));
-                    Title = "Star or Artifact";
-                    Description = "Choose \"Star or Artifact\" if there is no central galaxy to classify or if the artifact is so badly displayed that you can't ignore it and classify the galaxy with reasonable confidence.";
-                    Description2 = "The telescopes taking our data were designed to look at faint galaxies, which means that bright, compact objects like stars sometimes look a bit strange.";
-                    break;
+                    return new FeaturesGalaxy();
```
wgranger: These items were returned back to their factory settings so each example would have a unique class name, which is necessary when providing some of the animation logic in the individual styles under ExamplesPanel.xaml

GalaxyZooTouchTable/GalaxyZooTouchTable/Converters/ObjectToTypeStringConverter.cs
```diff
@@ -0,0 +1,23 @@
...
+            if (value != null)
+            {
+                return value.GetType().Name;
+            }
+            return null;
+        }
+
+        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            throw new NotImplementedException();
```
noahmalmed: Should this be implemented?

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/GalaxyExample.cs
noahmalmed: You couldn't just check the type value?

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/GalaxyExample.cs
```diff
@@ -8,47 +7,64 @@ public enum GalaxyType
...
+                case GalaxyType.NotAGalaxy:
+                    return new NotAGalaxy();
+                default:
```
noahmalmed: I don't know in C#, but is this default case required if your cases are exhaustive? If not, I think you should get rid of it.

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ExamplesPanelViewModel.cs
```diff
@@ -48,34 +52,34 @@ public GalaxyExample SelectedExample
...
             TogglePanel = new CustomCommand(SlidePanel, CanToggle);
-            SelectionChanged = new CustomCommand(SelectExample, CanChoose);
+            SelectItem = new CustomCommand(OnToggleItem, CanSelectItem);
+            UnselectItem = new CustomCommand(OnToggleItem);
         }
 
-        public void SelectExample(object sender)
+        public void OnToggleItem(object sender)
         {
-            if (sender == null)
```
noahmalmed: Is there a reason this is no longer needed?

GalaxyZooTouchTable/GalaxyZooTouchTable/Views/ExamplesPanel.xaml.cs
```diff
@@ -17,102 +11,11 @@ public partial class ExamplesPanel : UserControl
...
-        {
-            foreach (var example in ExampleList.Items)
-            {
-                ListBoxItem boxItem = (ListBoxItem)(ExampleList.ItemContainerGenerator.ContainerFromItem(example));
-
-                if (ExampleList.SelectedItem != example)
-                {
-                    AnimateToggleOpacity(boxItem, true);
-                }
-            }
```
noahmalmed: 👋 bye bye code! Where is all of this logic moved to?

GalaxyZooTouchTable/GalaxyZooTouchTable/Converters/ObjectToTypeStringConverter.cs
wgranger: It's unnecessary to have a `ConvertBack` method for this converter, but the interface requires a method to be defined. This is often the practice to have a `NotImplementedException` as a placeholder.

noahmalmed: I figured as much, just wanted to make sure you weren't leaving something out by accident.

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ExamplesPanelViewModel.cs
wgranger: Yes, this was originally used when a `ListView` element was used in place of all galaxy examples. Looks like I need to remove the `ICommand` definition above as well. 

`SelectItem` and `UnselectItem` now handle this logic as we are no longer using the finicky `ListView`.

GalaxyZooTouchTable/GalaxyZooTouchTable/Views/ExamplesPanel.xaml.cs
wgranger: All of this logic is now magically handled through the animations defined in the `Resources` section of `ExamplesPanel.xaml`

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/GalaxyExample.cs
wgranger: I'll look into this. I agree, it would be preferable to remove the default. I seem to remember C# complains without a `default`, but I might be mistaken.

GalaxyZooTouchTable/GalaxyZooTouchTable/Views/ExamplesPanel.xaml.cs
noahmalmed: ✨ ✨

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/GalaxyExample.cs
wgranger: Looks like I'll either have to return a default or return `null` after the switch statement. It might make the most sense here to throw an exception "Didn't find value," etc. if the default is reached.

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/GalaxyExample.cs
wgranger: I think this is the cleanest way to check with the animations on ExamplesPanel.xaml. The converter being used checks for the name of each object. Previously the name for all three objects was `GalaxyExample`

wgranger:
This is ready for another review.

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/GalaxyExample.cs
noahmalmed: fair enough!

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/GalaxyExample.cs
noahmalmed: I think it's fine! Just though it would make it easier to spot issues. I don't think it's worth changing anything

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/GalaxyExample.cs
```diff
@@ -33,7 +33,7 @@ public static GalaxyExample Create(GalaxyType type)
                 case GalaxyType.NotAGalaxy:
                     return new NotAGalaxy();
                 default:
-                    return new NotAGalaxy();
+                    throw new Exception("Example model not found.");
```
noahmalmed: 👍

noahmalmed:
Will test this tomorrow

noahmalmed:
Some notes from testing:

- Once the example panel is pulled out, I have a lot trouble getting it to go back in.
- It seems like the text in the answer buttons is incorrect. It says "Smooth_Gradually fades from the center_". Are those underscores supposed to be there?
- It is really difficult in general to get anything in the example bar to work on the first tap. I'm not sure if that's an issue with my device or WPF, but its taking multiple taps to do anything

wgranger:
I'll take a look at point one and three above. It might be that I need to update some of the touch responses. 

As for the second point, the text currently looks strange because it is dependent on a change in #19 that recognizes italicized notation in Markup.

wgranger:
This should be ready for another review. I enlarged the hit area for selecting an item on the examples pullout. In doing this, I had to add a `HandlingEventTrigger` component so commands would not tunnel. This was necessary because if you unselected an item, a second command would also fire on the parent component to select the item again, so the event needed to be marked as handled after the first event was met.

noahmalmed:
merge conflicts from merge of other PR

wgranger:
Merge conflict taken care of.

wgranger:
Are any changes needed to this code?

noahmalmed:
No. Sorry, tested Monday and it looks good to me.

PR #24, opened 2018-11-13
Optimize Sizes

wgranger:
This is the first of a few PRs to optimize the app for the large screen. We identified several items that need changing, but I thought it best to break these changes up over a few PRs to keep commits down.

Aside from sizing issues, I made a change here to the config file as there was a bug with the files not importing values before the client was making calls to Panoptes (thus crashing the app). A quick fix is to directly inject the values into the config here, as we won't need to be making use to production any time soon.


wgranger:
This branch is also installed on the touch table if you would like to test it out.

noahmalmed:
Hard to review the code because it is mostly XML changes. Can you list the things that are supposed to be fixed so I can verify them when I test the code?

wgranger:
This PR changes several things on multiple views. 

General:
- Remove cursors from the app
- Add "ask a friend" placeholders
- Fix bug with galaxy examples remaining in expanded-size when example is selected and classifier closed, then reopened
- Use higher-res logo and Uscientist images

Centerpiece:
- Change font to Poppins
- Widen the centerpiece a smidge

Subject Viewer:
- Adjust corner radius to answer buttons
- Adjust corner radius of white subject border
- Adjust margins of green/yellow images on answer buttons

Classification Panel:
- Use Font Awesome image for "close confirm" X
- Adjust layout of green submit button
- "Classify another galaxy" on green button needed Font Awesome chevron sign

Summary View
- Fix awkward margin between number of classifications received out of total. Eg. "(3/25)"
- Adjust grid margins on "Nice work... to study" header text
- Adjust layout of progress bar results from past classifications
- Disable slight bug where user could select progress bar, thus turning it green
- "UScientist" database should be "Zooniverse" database

Start Button
- "Tap here" text should be smaller and correctly colored, not white
- Use Poppins font

Close Confirmation
- Don't capitalize "Zooniverse" on url
- Blur drop shadow a bit
- Resize button text for "close" and "continue"
- Place "iOS" on same line as "Android..."
- Adjust overall grid margins

Leveler
- No caps on "More"

wgranger:
Also note, anywhere containing a margin definition, `Margin="5"` is generally bad for keeping the app responsive, laptop -> touch table.

noahmalmed:
Thanks @wgranger, that's very verbose! I'll verify the changes tomorrow.

noahmalmed:
Some notes 
- Is the ask a friend place holder just supposed to be a black box?
- Should the examples stay expanded when the example view is hidden and then shown?

Let me know about these ^^. Otherwise, everything else looks good

wgranger:
Yep! The placeholder is just a black box for now. The examples should stay expanded. There may be a bug with some of the text being visible on the tab when it's closed back in, but that's fixed in the current PR that is open.

noahmalmed:
Looks good then. Merging

PR #30, opened 2018-11-30
Use Exact Sizes + Design Updates

### LINKED ISSUE #3: Size based on view size not screen size
Currently the app sizes itself based off the size of the screen rather than the size of the view.

Not sure if it's a big deal, but it may be a good idea to size off the size of the view

### LINKED ISSUE #4: Touch only shows feedback for one
When multiple touches are happening the app only show touch feedback for one of the touches

### LINKED ISSUE #21: Update Display for Touch Table
Now that some of the publishing issues are being taken care of, there are some screen size issues that need to be taken care of when the app is open on a larger screen (touch table vs. laptop view).

wgranger:
This PR was actually not as large as I thought it would be. However, there are many fixes here. Basically, we can abide by more absolute numbers using the `Viewbox` if we keep the screen in a 16:9 aspect ratio, which is what the table is at. This cuts down on some messy code and makes the design look more crisp. There should be one more event handler that needs adjusting, but it will take more effort and I'll open an issue about it now.

This version is currently installed on the touch table.

Closes #21 
Closes #4
Closes #3


GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ClassificationPanelViewModel.cs
```diff
@@ -314,9 +302,28 @@ private async void GraphQLRequest()
...
+
             ResetAnswerCount();
+            GraphQLResponse response = new GraphQLResponse();
+
+            try {
+                response = await GraphQLClient.SendQueryAsync(answersRequest);
+            }
+            catch (Exception e)
+            {
+                Console.WriteLine("Graph QL Error: {0}", e.Message);
```
noahmalmed: Will there eventually be any error handling beyond writing to the console?

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/MainWindowViewModel.cs
```diff
@@ -2,10 +2,12 @@
 using GalaxyZooTouchTable.Utility;
 using PanoptesNetClient;
 using PanoptesNetClient.Models;
+using System;
```
noahmalmed: You should probably avoid importing all of `System`

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ClassificationPanelViewModel.cs
wgranger: Yes, there probably should be some. It would be good to find an external service, whether that is Rollbar or something else. I'll open an issue about it, because it would be good to automate some of this to keep a keen eye on the table.

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/MainWindowViewModel.cs
wgranger: I don't think importing all of `System` affects performance, but I've removed it for posterity.

noahmalmed:
@wgranger things looked better on the table. It looks like the answer button subtexts are still kind of blurry. May just be from their text size and the resolution of the screen though. Let me know what you think about that

noahmalmed:
@wgranger things looked better on the table. It looks like the answer button subtexts are still kind of blurry. May just be from their text size and the resolution of the screen though. Let me know what you think about that

wgranger:
I believe thats due to the resolution and screen. However, I'm not currently worried about it as the answer button component will change slightly in the future (from `ListView` to `ItemsControl`) to allow for better touch handling.

noahmalmed:
lgtm then

PR #33, opened 2018-12-06
Ask a Friend Components and Functionality

### LINKED ISSUE #25: Introduce Ask a Friend
Turns out we will be introducing this in Phase One. Functionality available via https://app.atomic.io/d/NVEgDvEukdAY

wgranger:
Closes #25 

Should probably wait until #30 is merged, as a few UI tweaks will likely be needed here.


wgranger:
Recently rebased and currently on the touch table.

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ClassificationPanelViewModel.cs
```diff
@@ -216,6 +230,12 @@ private async void OnContinueClassification(object sender)
                 ClassificationsThisSession += 1;
                 Messenger.Default.Send<int>(ClassificationsThisSession, User);
                 CurrentView = SUMMARY_VIEW;
+
+                if (Notifications.Status == NotificationStatus.HelpingUser)
+                {
+                    Notifications.SendAnswerToUser(SelectedItem);
+                }
+                Notifications.OnResetNotifications(null);
```
noahmalmed: Seems weird that you are passing `null` here. Do you maybe want to pass `this`?

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/MainWindowViewModel.cs
```diff
@@ -110,7 +110,39 @@ public MainWindowViewModel()
             CreateTimer();
             LoadCommands();
 
-            ActiveUsers.CollectionChanged += ActiveUsersChanged;
+            AllUsers.CollectionChanged += AllUsersCollectionChanged;
+        }
+
+        private void AllUsersCollectionChanged(object sender, NotifyCollectionChangedEventArgs e)
```
noahmalmed: Could you use a more descriptive var name than `e`? Maybe something like `changedEventArgs`?

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/MainWindowViewModel.cs
```diff
@@ -110,7 +110,39 @@ public MainWindowViewModel()
...
+            if (e.OldItems != null)
+            {
+                foreach (object item in e.OldItems)
+                {
+                    ((INotifyPropertyChanged)item).PropertyChanged -= ItemPropertyChanged;
+                }
+            }
+        }
+
+        private void ItemPropertyChanged(object sender, PropertyChangedEventArgs e)
```
noahmalmed: Same here re: `e`

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/NotificationsViewModel.cs
```diff
@@ -0,0 +1,207 @@
...
+            {
+                _user = value;
+                OnPropertyRaised("User");
+            }
+        }
+
+        public NotificationsViewModel(TableUser user, ObservableCollection<TableUser> allUsers, ClassificationPanelViewModel classifier)
+        {
+            Classifier = classifier;
+            user.Notifications = this;
```
noahmalmed: Can you go into more detail on why the user this class both need references to each other. 

Also, I think it's a little confusing that the constructor for `NotificationsViewModel` modifies a value of one of the parameters passed into is. Generally, I wouldn't expect a constructor to have this kind of side effect.

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ClassificationPanelViewModel.cs
wgranger: I admit that is rather odd. `onResetNotifications` requires an argument for the `CustomCommand` class. I might be able to set the default param to `null` on the method itself. If not, changing `null` to `this` should do the trick here.

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/NotificationsViewModel.cs
wgranger: Yeah, this is a bit iffy I'll admit, and it does create a dependency that might be avoidable. It's tricky with the notifier needing to send a message to the classifier and user. Let me see if I can use the Messenger object to take care of this functionality.

wgranger:
Several changes in the past few commits:

- Cleaned up dependencies between user and notifications view model. Now each user has status attached directly to the user with a separate notification view model on the classifier. The messenger helped a bit with this. Several lines were added, but the code reads a bit cleaner too.

- Made some design updates: introduced an hourglass icon when user is busy and tweaked appearance of several components.

- Correctly clear out notifications if your peer leaves in the process of your communication.

The touch table has been updated to reflect these changes.

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ClassificationPanelViewModel.cs
```diff
@@ -224,7 +224,7 @@ private async void OnContinueClassification(object sender)
                 CurrentClassification.Metadata.FinishedAt = System.DateTime.Now.ToString();
                 CurrentClassification.Annotations.Add(CurrentAnnotation);
                 ApiClient client = new ApiClient();
-                await client.Classifications.Create(CurrentClassification);
+                //await client.Classifications.Create(CurrentClassification);
```
noahmalmed: can you get rid of this?

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/TableUserFactory.cs
```diff
@@ -17,18 +17,19 @@ enum UserType
 
     public class TableUser : INotifyPropertyChanged
     {
+        public virtual string Name { get; }
```
noahmalmed: I like that you are using virtual for this property, but why aren't you using it for the other properties in this class?

noahmalmed: Why does it need to be overridden? It seems comparable to `ThemeColor`, `StartButton` and `Avatar`

wgranger: Without the override, `Name` takes the `null` value of the base class (`TableUser`) and will not work with the messenger. Take a look at the use of the `Messenger` in this PR. The name property is used there to create unique values for the messenger dictionary.

noahmalmed: But you only ever make TableUser (inheriting) objects through the factory, right? So as long as  `Name` is defined in the object produced by the factory method then I don't see how `Name` would ever be null. How is `name` used any differently than `ThemeColor`?

wgranger: `ThemeColor` is referenced through bindings in views, so the property is correctly retrieved from the object (whether `FaceUser` or `PersonUser`). However, take a look at the `NotificationsViewModel`. That model only knows that the User is a `TableUser`, not the type. So it still looks at the base class for definitions.

noahmalmed: Huh, maybe I'm just confused by how C# sharp works. It may be worth making the other props in `TableUser` virtual as well just in case you may have to use those values in code in the future. Also, I think it makes it more explicit that those values aren't supposed to be defined on the `TableUser` and that `TableUser` should really just be a virtual class.

wgranger: In that case, I should probably revert back to marking `TableUser` as an abstract class, which is more aligned with what I want.  I'll push up a fix momentarily.

wgranger:
@noahmalmed Go ahead and take another look at this. I'm unable to recreate the bug I experienced earlier, but try to break it if you can. I'll likely have to make some changes to this component later as we implement more design changes, so if you can't find any bugs with the notifier, there's a chance future changes would take care of any issues.

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/TableUserFactory.cs
```diff
@@ -15,12 +15,12 @@ enum UserType
         Heart
     }
 
-    public class TableUser : INotifyPropertyChanged
+    public abstract class TableUser : INotifyPropertyChanged
```
noahmalmed: I'm glad this is an abstract class. It's makes the most sense for this class bc it should never be instantiated and this will help enforce that.

noahmalmed:
Not a merge blocking comment, but should open an issue on it. 

You can circularly classify the same galaxy between two users by doing the following.

- User 1) Classify galaxy
- User 1) [On science'd screen] ask for user 2's help
- User 2) Accept User 1 invitation
- User 2) classify galaxy
- User 1) [Still on science'd screen] close side notification window and close overlay over the ask neighbor view.
- User 2) [Still on science'd screen] Ask User 1's help for same galaxy

noahmalmed:
Besides the issue above, gonna merge

PR #35, opened 2018-12-20
Fix Hit State by Using ItemsControl

### LINKED ISSUE #31: Fix Event Handler on Answer Buttons
Currently, all commands have been attached to the `touchUp` command to allow multi events at the same time. However, the answer buttons will remain unresponsive in a touch is currently occurring on the board. 

This is seem if you hold a finger anywhere on the board and try to interact with any of the answer buttons.

This will require a bit more thinking since the buttons are contained in a `Listview` and annotations are tied to the `IsSelected` property of the list view.

wgranger:
Closes #31 

The solution to this was simple: just use an `ItemsControl` instead of a `ListView`. Note the use of the EqualityConverter here, which is the same as the one in #33.


wgranger:
@noahmalmed This has been rebased and is good to go.

noahmalmed:
works well. merging

PR #40, opened 2019-01-07
Remove dependencies

### LINKED ISSUE #39: Separation of Concerns
There's a bit more work that needs to be done before tests can be implemented. Mainly, there needs to be a better separation of concerns in  the app. This includes removing dependencies in view models and working towards better encapsulation. 

A general reorganization of some view models should take care of this.

wgranger:
This is ready for an initial review, but I'd like to get #36 merged first so I can add some changes in that branch to this PR. 

A good bit of the structure was changed here, mainly to remove view model dependencies in order to begin implementing tests. Some more minor structural changes might be necessary when tests are added, but the major changes are reflected here.

**Note**: I removed the `TableUserFactory` as a similar factory was needed here when constructing each `ClassificationPanelViewModel`. 

Closes #39


wgranger:
This should not be good for a review. Please let me know if anything here is confusing, as I made a good bit of structural changes. @noahmalmed

noahmalmed:
not or now?

wgranger:
Oops, *now

GalaxyZooTouchTable/GalaxyZooTouchTable/Lib/CommonData.cs
```diff
@@ -0,0 +1,50 @@
+﻿using GalaxyZooTouchTable.Models;
+using System;
+using System.Collections.ObjectModel;
+using System.ComponentModel;
+
+namespace GalaxyZooTouchTable.Lib
+{
+    public class CommonData : INotifyPropertyChanged
```
noahmalmed: The name and purpose of this class is kind of confusing. Does it just contain all the user classes?

Could you change the name of the class or add a comment to better explain the purpose of the class.

GalaxyZooTouchTable/GalaxyZooTouchTable/Lib/CommonData.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            if (_instance == null)
+                _instance = new CommonData();
+
+            return _instance;
+        }
+
+        private void PopulateUsers()
+        {
+            AllUsers.Add(PersonUser);
```
noahmalmed: Instead of having the user classes be member variables, it may make more sense to just initialize and add them to AllUsers here like: 

`AllUsers.Add(new PersonUser());`

I think you should do this if you aren't ever planning on referencing the individual user classes

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/ClassificationPanelViewModelFactory.cs
```diff
@@ -0,0 +1,45 @@
...
+            if (panoptesRepo == null || graphQLRepo == null)
+            {
+                throw new ArgumentNullException("RepoDependency");
+            }
+
+            _panoptesRepo = panoptesRepo;
+            _graphQLRepo = graphQLRepo;
+        }
+
+        public ClassificationPanelViewModel Create(UserType type)
```
noahmalmed: 👍 I like this refactor

GalaxyZooTouchTable/GalaxyZooTouchTable/Services/PanoptesRepository.cs
```diff
@@ -0,0 +1,33 @@
+﻿using PanoptesNetClient;
+using PanoptesNetClient.Models;
+using System.Collections.Generic;
+using System.Collections.Specialized;
+using System.Threading.Tasks;
+
+namespace GalaxyZooTouchTable.Services
+{
+    public class PanoptesRepository : IPanoptesRepository
```
noahmalmed: `Repository` seems like a weird word to use here. Why not call this `PanoptesService` to be more consistent with your directory naming structure

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/CenterpieceViewModel.cs
```diff
@@ -0,0 +1,87 @@
...
+                foreach (object item in changedEventArgs.OldItems)
+                {
+                    ((INotifyPropertyChanged)item).PropertyChanged -= ItemPropertyChanged;
+                }
+            }
+        }
+
+        private void ItemPropertyChanged(object sender, PropertyChangedEventArgs changedEventArgs)
+        {
+            foreach (TableUser user in AllUsers)
```
noahmalmed: You may want to use `find` here to save some lines https://docs.microsoft.com/en-us/dotnet/api/system.array.find?view=netframework-4.7.2.

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/CenterpieceViewModel.cs
```diff
@@ -0,0 +1,87 @@
...
+
+        private bool _showJoinMessage = true;
+        public bool ShowJoinMessage
+        {
+            get => _showJoinMessage;
+            set => SetProperty(ref _showJoinMessage, value);
+        }
+
+        private bool _flipCenterpiece;
+        public bool FlipCenterpiece
```
noahmalmed: I think this var should be called something like `CenterPieceIsFlipped`. `FlipCenterpiece` sounds more like a command.

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ClassificationPanelViewModel.cs
```diff
@@ -119,27 +116,50 @@ public AnswerButton SelectedAnswer
...
+            Notifications.SendRequestToUser += OnSendRequestToUser;
+            StillThere.ResetFiveMinuteTimer += ResetTimer;
+            StillThere.CloseClassificationPanel += CloseClassificationPanel;
+        }
 
-            if (CurrentTask.Answers != null)
-            {
-                CurrentAnswers = ParseTaskAnswers(CurrentTask.Answers);
-            }
+        private void CloseClassificationPanel()
```
noahmalmed: Seems like you can just get rid of this function if all it does is just call another function

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/StillThereViewModel.cs
```diff
@@ -112,7 +111,7 @@ private void OneSecondElapsed(object sender, System.EventArgs e)
             CurrentSeconds--;
             decimal StartingSeconds = 30;
             decimal PercentOfSeconds = (CurrentSeconds / StartingSeconds) * 100;
-            Percentage = Convert.ToInt16(Math.Floor(PercentOfSeconds));
+            Percentage = System.Convert.ToInt16(System.Math.Floor(PercentOfSeconds));
```
noahmalmed: Why did you remove the `System` import?

GalaxyZooTouchTable/GalaxyZooTouchTable/Lib/CommonData.cs
wgranger: `GlobalData` might be a better title here.

GalaxyZooTouchTable/GalaxyZooTouchTable/Lib/CommonData.cs
wgranger: I have it set up this way because the PersonUser added to `AllUsers` should be the same as the one that is created when `ClassificationPanelViewModel` is instantiated with a PersonUser. 

Because of this, I create the global PersonUser first in this constructor, and refer to that instance for the addition to `AllUsers` as well as the `CommonData.GetInstance().PersonUser` that is passed to the `ClassificationPanelViewModel` constructor. This is necessary when doing some comparison of users in the `NotificationsViewModel`.

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/ClassificationPanelViewModelFactory.cs
wgranger: Yes, this was originally what prompted this PR and is necessary to get testing wired up.

GalaxyZooTouchTable/GalaxyZooTouchTable/Services/PanoptesRepository.cs
noahmalmed: https://stackoverflow.com/questions/5049363/difference-between-repository-and-service-layer

Did some research. Looks like the `Repository` naming convention is for basic CRUD operations. According to this post, I think this class is more business logicky and would make more sense being named `Service`

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/StillThereViewModel.cs
noahmalmed: Yeah, sorry don't mean to be flip flopping. Can you import `System.Convert`?

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/CenterpieceViewModel.cs
wgranger: `find` isn't available on an ObservableCollection, but I can shorten these lines by using LINQ

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/StillThereViewModel.cs
wgranger: Nah. That's unavailable as a sole import. There are also other items on this view model using `System` eg. `System.Action`.

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ClassificationPanelViewModel.cs
wgranger: Oh duh. This was expecting an `object sender` to be passed through the delegate.

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/StillThereViewModel.cs
wgranger: I wound up importing `system` since it was used throughout this component, but I'm open to changing this is needed.

wgranger:
This is ready for another review.

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/CenterpieceViewModel.cs
```diff
@@ -60,15 +61,7 @@ private void AllUsersCollectionChanged(object sender, NotifyCollectionChangedEve
...
-            foreach (TableUser user in AllUsers)
-            {
-                if (user.Active)
-                {
-                    ShowJoinMessage = false;
-                    return;
-                }
-            }
-            ShowJoinMessage = true;
+            ShowJoinMessage = !AllUsers.Any(user => user.Active == true);
```
noahmalmed: nice

PR #41, opened 2019-01-17
Add Space Centerpiece

wgranger:
This PR adds functionality for dragging and dropping a galaxy onto a classifier. It also provides some math for putting galaxies in their correct places on a cutout. Although there is some hard-coded variables here, that will change once this is fleshed out more and users can navigate to other cutouts with a subject set that actually contains galaxies in those coordinate (RA/DEC) bounds.

**A few notes here**:
- WPF has built in drag and drop functionality that works with drag handlers such as `DragOver`, `DragEnter`, etc. but that is triggered with `DragOver.DoDragDrop`, which blocks the UI and won't register touches through the duration of the drag.

- Apparently there was once a magic Surface SDK (PixelSense) to handle multi-touch drag that Microsoft maintained, but they've mysteriously deleted all docs from their site, which tells me that SDK is no longer maintained.

Hence the solution here to support multi-touch. Some functionality was also needed here to block classifier UI elements if a subject isn't present (selecting an answer, asking a neighbor, etc.).


GalaxyZooTouchTable/GalaxyZooTouchTable/Models/TableSubject.cs
```diff
@@ -0,0 +1,25 @@
...
+using System;
+
+namespace GalaxyZooTouchTable.Models
+{
+    public class TableSubject
+    {
+        public string X { get; set; }
+        public string Y { get; set; }
+        public double RA { get; set; }
+        public double DEC { get; set; }
```
noahmalmed: Can you choose better names for `RA` and `DEC`? It's sort of confusing what they are

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/TableSubject.cs
```diff
@@ -0,0 +1,25 @@
...
+        public string SubjectLocation { get; set; }
+        private double PlateScale { get; set; } = 0.4;
+        public Subject Subject { get; set; }
+
+        public TableSubject(Subject subject, double tableRA, double tableDec)
```
noahmalmed: Same here with `tableRA` and `tableDec`

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/SpaceViewModel.cs
```diff
@@ -0,0 +1,61 @@
...
+            }
+        }
+
+        public SpaceViewModel()
+        {
+            GetSpaceCutout();
+            GetSubjectsAsync();
+        }
+
+        private void GetSpaceCutout()
```
noahmalmed: Would it make more sense to just move the content of this function inline with the constructor?

Seems overly verbose to have a whole function for just one line of code

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/SpaceViewModel.cs
```diff
@@ -0,0 +1,61 @@
...
+using System.ComponentModel;
+using System.Threading.Tasks;
+
+namespace GalaxyZooTouchTable.ViewModels
+{
+    public class SpaceViewModel : INotifyPropertyChanged
+    {
+        public List<TableSubject> CurrentGalaxies { get; set; } = new List<TableSubject>();
+        public double RA { get; set; } = 250.2;
+        public double DEC { get; set; } = 35.1;
```
noahmalmed: Are these the hardcoded values you mentioned in the PR intro?

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/TableSubject.cs
```diff
@@ -5,21 +5,71 @@ namespace GalaxyZooTouchTable.Models
...
             SubjectLocation = subject.GetSubjectLocation();
+            EQConvert(TableRA, TableDEC);
+        }
+
+        private void EQConvert(double TableRA, double TableDEC)
+        {
+            double PI = 3.1415926535897932384626433832795;
+            double TwoPI = 6.283185307179586476925286766559;
+            double PI2 = 1.5707963267948966192313216916398;
+            double Rads = 0.01745329251994329576923690768489;
```
noahmalmed: Does C# not ship with these constants? Looks like at least pi is available here: https://docs.microsoft.com/en-us/dotnet/api/system.math?view=netframework-4.7.2 ?

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/TableSubject.cs
```diff
@@ -5,21 +5,71 @@ namespace GalaxyZooTouchTable.Models
...
+
+            double Az, Alt, tmpval, nz, az2, tx, ty, xyscale;
+
+            double DisplayFOV = 1.4;
+            double DisplayROT = 0.0;
```
noahmalmed: I know there is a lot of math here and the `FOV` and `ROT` probably stand for some math stuff, but I think it would make sense to either rename or add a large comment explaining what these values mean.

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/TableSubject.cs
```diff
@@ -5,21 +5,71 @@ namespace GalaxyZooTouchTable.Models
...
+        }
+
+        private void EQConvert(double TableRA, double TableDEC)
+        {
+            double PI = 3.1415926535897932384626433832795;
+            double TwoPI = 6.283185307179586476925286766559;
+            double PI2 = 1.5707963267948966192313216916398;
+            double Rads = 0.01745329251994329576923690768489;
+
+            double Az, Alt, tmpval, nz, az2, tx, ty, xyscale;
```
noahmalmed: As well, a lot of these names aren't very descriptive.

GalaxyZooTouchTable/GalaxyZooTouchTable/Behaviors/UIElementDragBehavior.cs
```diff
@@ -0,0 +1,124 @@
...
+        {
+            if (DragOverlay == null)
+            {
+                var visual = e.OriginalSource as Visual;
+                DragOverlay = (DragCanvas)UIElementDragBehavior.FindAncestor(typeof(DragCanvas), visual);
+            }
+
+            if (isTouchDown)
+            {
+                TouchPoint touchPosition = e.GetTouchPoint(DragOverlay);
```
noahmalmed: For clarity purposes, it may be useful to pull this code into a function with a descriptive title. As is, it's sort of confusing what is happening here.

GalaxyZooTouchTable/GalaxyZooTouchTable/Behaviors/UIElementDragBehavior.cs
```diff
@@ -0,0 +1,124 @@
...
+
+        private void DragDropContainer_TouchUp(object sender, TouchEventArgs e, GalaxyAdorner adorner, UIElement element)
+        {
+            var touchPosition = e.GetTouchPoint(DragOverlay);
+            Point touchPoint = new Point(touchPosition.Position.X, touchPosition.Position.Y);
+
+            if (touchPoint.X >= adorner.Location.X && touchPoint.X <= adorner.Location.X + adorner.ActualWidth
+                && touchPoint.Y >= adorner.Location.Y && touchPoint.Y <= adorner.Location.Y + adorner.ActualHeight)
+            {
+                AdornerLayer.GetAdornerLayer(sender as Visual).Remove(adorner);
```
noahmalmed: Same thing here re: pulling into a function

GalaxyZooTouchTable/GalaxyZooTouchTable/Behaviors/UIElementDragBehavior.cs
```diff
@@ -0,0 +1,124 @@
...
+            var adornerLocation = new Point(touchPosition.Position.X, touchPosition.Position.Y);
+            adorner.UpdatePosition(adornerLocation);
+        }
+
+        private void DragDropContainer_TouchUp(object sender, TouchEventArgs e, GalaxyAdorner adorner, UIElement element)
+        {
+            var touchPosition = e.GetTouchPoint(DragOverlay);
+            Point touchPoint = new Point(touchPosition.Position.X, touchPosition.Position.Y);
+
+            if (touchPoint.X >= adorner.Location.X && touchPoint.X <= adorner.Location.X + adorner.ActualWidth
```
noahmalmed: It may make sense to assign this boolean check to a variable just to better describe what you are checking

GalaxyZooTouchTable/GalaxyZooTouchTable/Behaviors/UIElementDropBehavior.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        private void AssociatedObject_TouchEnter(object sender, TouchEventArgs e)
+        {
+            Border element = sender as Border;
+            LinearGradientBrush brush = new LinearGradientBrush();
+            brush.StartPoint = new Point(0, 0.5);
+            brush.EndPoint = new Point(1, 0.5);
+            brush.GradientStops.Add(new GradientStop((Color)ColorConverter.ConvertFromString("#727272"), 0));
+            brush.GradientStops.Add(new GradientStop((Color)ColorConverter.ConvertFromString("#202020"), 1));
+            brush.GradientStops.Add(new GradientStop((Color)ColorConverter.ConvertFromString("#000000"), 1));
```
noahmalmed: Are the color constants here used throughout the app? It may at least be worth grabbing the color constant for black from the c sharp color library: https://docs.microsoft.com/en-us/dotnet/api/system.drawing.color.black?view=netframework-4.7.2#System_Drawing_Color_Black

GalaxyZooTouchTable/GalaxyZooTouchTable/Behaviors/UIElementDropBehavior.cs
```diff
@@ -0,0 +1,55 @@
...
+            element.Background = brush;
+
+            Border childBorder = element.Child as Border;
+            childBorder.Background = (SolidColorBrush)(new BrushConverter().ConvertFrom("#4DE5FF4D"));
+            DropShadowEffect childEffect = new DropShadowEffect();
+            childEffect.BlurRadius = 20;
+            childEffect.Color = Color.FromArgb(120,255,255,255);
```
noahmalmed: I think it would probably be best if you stuck to one way of defining a color. Either the `ColorConverter` used above or the `Color` you use here. Personally, I think the `Color` lib looks better

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/TableSubject.cs
```diff
@@ -0,0 +1,45 @@
...
+            DEC = System.Convert.ToDouble(subject.Metadata.dec);
+            Subject = subject;
+            SubjectLocation = subject.GetSubjectLocation();
+            XYConvert(TableRA, TableDEC);
+        }
+
+        private void XYConvert(double TableRA, double TableDEC)
+        {
+            int ScreenWidth = 1248; 
+            int ScreenHeight = 432;
```
noahmalmed: Are these values not available to you through the code?

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/TableSubject.cs
```diff
@@ -0,0 +1,45 @@
...
+            XYConvert(TableRA, TableDEC);
+        }
+
+        private void XYConvert(double TableRA, double TableDEC)
+        {
+            int ScreenWidth = 1248; 
+            int ScreenHeight = 432; 
+
+            double DecRange = ScreenHeight * PlateScale / 3600;
+            double RaRange = System.Math.Abs(ScreenWidth * PlateScale / 3600 / System.Math.Cos(TableDEC));
```
noahmalmed: Curious what the `3600` number is for

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/SpaceViewModel.cs
```diff
@@ -0,0 +1,61 @@
...
+using System.Collections.Specialized;
+using System.Threading.Tasks;
+
+namespace GalaxyZooTouchTable.ViewModels
+{
+    public class SpaceViewModel : ViewModelBase
+    {
+        public double RA { get; set; } = 250.3035;
+        public double DEC { get; set; } = 35.09;
+        public double SCALE { get; set; } = 1.5;
```
noahmalmed: Are these the hardcoded values you were referring to in the description of the PR?

noahmalmed:
Dragable galaxies are not showing up on my build

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/TableSubject.cs
wgranger: Those are pretty standard names for astrometry, but I can do RightAscension and Declination

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/SpaceViewModel.cs
wgranger: As the centerpiece is built out, this method will have to be called whenever the space navigation changes to a different tile. This method will return a different image as the RA/DEC are changed. Note the string interpolation in this method.

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/SpaceViewModel.cs
wgranger: Yes, these will change in the future, but they are hardcoded now because I'm working with a test subject set that contains subjects within these specific RA/DEC bounds.

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/TableSubject.cs
wgranger: Ahh, yes. Good catch. I was working with a few different solutions here and forgot to clean up completely.

wgranger: Actually, this was cleaned up already. This is just an outdated comment.

GalaxyZooTouchTable/GalaxyZooTouchTable/Behaviors/UIElementDropBehavior.cs
wgranger: `Color.FromArgb` will still be necessary for colors that need an opacity or aren't defined in the `Colors.COLOR_NAME` namespace.

wgranger: Also, some of these properties require the `SolidColorBrush` namespace, while others want the `Color` namespace, which is why there's a difference.

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/TableSubject.cs
wgranger: They are not. There is a separation of concerns between the Views and the Viewmodels. Each `TableSubject` is constructed in the `SpaceViewModel`, but we need to know the size of the cutout width and height to correctly find out the pixel placement of each subject. 

That said, these variables should be changed to `CutoutWidth` and `CutoutHeight` to more accurately reflect what they represent.

GalaxyZooTouchTable/GalaxyZooTouchTable/Behaviors/UIElementDropBehavior.cs
noahmalmed: I see. That's an unfortunate code pattern, but looks like it's what you have to do. 👍

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/TableSubject.cs
noahmalmed: Even those may be standard names for astrometry, the people working on the code may not be familiar with the terms. So I think it's best to be verbose with language here.

noahmalmed:
I don't think these should block merge, but I just wanted to make a couple of notes. 

1. Perhaps the 'move map' buttons should be outside of the draggable view in case a galaxy location overlaps with them.
2. If a user drags a galaxy to their view after already having a galaxy up, perhaps it should reset their workstation with that galaxy

PR #42, opened 2019-01-23
Fix Minor Bugs

wgranger:
This PR fixes a few minor bugs we noticed when testing the touch table on the floor.

- Resets the leveler count when the classifier closes, which wasn't the previous behavior.
- Decreases time to see the "Still There?" modal from 5 to 3 minutes.
- On the classification summary view, sets a minimum width for a progress bar. A bug can occur when there are 25 classifications for a subject, but one answer only has one vote. However, most of the work to prevent this from happening will have to be set up in Caesar/Panoptes with retirement rules to prevent more than 25 classifications from submitting.
- Increases the hit area to toggle off an example on the right examples panel.
- Changes the wording and animations of the header for each example on the examples panel.

### DIFF
--- a/GalaxyZooTouchTable/GalaxyZooTouchTable/Behaviors/DisableOpacityMaskOnEndScroll.cs
+++ b/GalaxyZooTouchTable/GalaxyZooTouchTable/Behaviors/DisableOpacityMaskOnEndScroll.cs
@@ -21,8 +21,8 @@ namespace GalaxyZooTouchTable.Behaviors
         private void SetOpacityMask()
         {
             var gradientCollection = new GradientStopCollection();
-            var gradientStop1 = new GradientStop(Color.FromRgb(0, 0, 0), 0);
-            var gradientStop2 = new GradientStop(Color.FromArgb(0, 0, 0, 0), 1);
+            var gradientStop1 = new GradientStop(Colors.Black, 0);
+            var gradientStop2 = new GradientStop(Colors.Transparent, 1);
             gradientCollection.Add(gradientStop1);
             gradientCollection.Add(gradientStop2);
             
@@ -44,7 +44,7 @@ namespace GalaxyZooTouchTable.Behaviors
             AssociatedObject.OpacityMask = InitialOpacityMask;
         }
 
-        public static readonly DependencyProperty IsHorizontalProperty =
+        private static readonly DependencyProperty IsHorizontalProperty =
             DependencyProperty.Register("IsHorizontal", typeof(Boolean), typeof(DisableOpacityMaskOnEndScroll));
 
         public bool IsHorizontal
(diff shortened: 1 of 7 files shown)


GalaxyZooTouchTable/GalaxyZooTouchTable/Behaviors/DisableOpacityMaskOnEndScroll.cs
```diff
@@ -0,0 +1,81 @@
...
+
+            AssociatedObject.ScrollChanged += AssociatedObject_ScrollChanged;
+            SetOpacityMask();
+        }
+
+        private void SetOpacityMask()
+        {
+            var gradientCollection = new GradientStopCollection();
+            var gradientStop1 = new GradientStop(Color.FromRgb(0, 0, 0), 0);
+            var gradientStop2 = new GradientStop(Color.FromArgb(0, 0, 0, 0), 1);
```
noahmalmed: Does `Color` ship with constants like `transparent` and `black`? It may be better to use those here instead of defining the colors with rgba values

GalaxyZooTouchTable/GalaxyZooTouchTable/Behaviors/DisableOpacityMaskOnEndScroll.cs
```diff
@@ -0,0 +1,81 @@
...
+        public bool IsHorizontal
+        {
+            get { return (bool)GetValue(IsHorizontalProperty); }
+            set { SetValue(IsHorizontalProperty, value); }
+        }
+
+        public static readonly DependencyProperty StartPercentProperty =
+            DependencyProperty.Register("StartPercent", typeof(double), typeof(DisableOpacityMaskOnEndScroll));
+
+        public double StartPercent
```
noahmalmed: Should these properties be `public`?

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/LevelerViewModel.cs
```diff
@@ -9,8 +9,12 @@ public class LevelerViewModel : ViewModelBase
         public TableUser User { get; set; }
         public ICommand ToggleLeveler { get; private set; }
         const string MAX_LEVEL = "Five";
+        const int DEFAULT_CLASSIFICATIONS_UNTIL_UPGRADE = 5;
+        const int DEFAULT_CLASSIFICATIONS_COUNT = 5;
+        const string DEFAULT_CLASSIFICATION_LEVEL = "One";
+        const bool DEFAULT_CLASSIFICATION_OPEN = false;
```
noahmalmed: Should these all be `private`?

GalaxyZooTouchTable/GalaxyZooTouchTable/Behaviors/DisableOpacityMaskOnEndScroll.cs
wgranger: Yes, the getter and setter need to be `public` so the views can pass in a property, but the `static readonly` value above can be `private`.

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/LevelerViewModel.cs
wgranger: They're implicitly `private`, but I can make it explicit.

noahmalmed:
conflicts

wgranger:
Fixed

noahmalmed:
looks good

PR #46, opened 2019-01-31
Add Tests To Project

### LINKED ISSUE #18: Add Tests
A test base should be added to the project.

wgranger:
This is the first addition of tests to the app. The tests here could be strengthened, but it will require more major changes that relies on reorganizing some of the code. I'll open a new issue of some organizational changes that are needed.

The main approach here is to introduce simple tests for each view model without the PR getting too large.

Closes #18


wgranger:
#47 has been created to track organization needs

noahmalmed:
conflicts

wgranger:
Fixed

noahmalmed:
Can you set these tests up to run with travis?

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/CenterpieceViewModel.cs
```diff
@@ -40,7 +41,7 @@ public CenterpieceViewModel()
             AllUsers.Add(GlobalData.GetInstance().EarthUser);
         }
 
-        private void AllUsersCollectionChanged(object sender, NotifyCollectionChangedEventArgs changedEventArgs)
+        public void AllUsersCollectionChanged(object sender, NotifyCollectionChangedEventArgs changedEventArgs)
```
noahmalmed: Instead of making all of these functions `public` for the sake of testing, it may be a better idea to find ways to trigger the functions they are meant to be triggered through existing public interfaces. 

It's not a huge deal, but I think it makes the code more readable to only publicly expose code that is meant to be publicly exposed

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ClassificationPanelViewModel.cs
```diff
@@ -168,16 +164,21 @@ private void AddSubscribers()
...
+        }
+
+        public void OnSendRequestToUser(TableUser UserToNotify)
         {
             NotificationRequest Request = new NotificationRequest(User, CurrentSubject.Id);
             Messenger.Default.Send<NotificationRequest>(Request, $"{UserToNotify.Name}_ReceivedNotification");
         }
 
-        private async void GetWorkflow()
+        public async Task GetWorkflow()
```
noahmalmed: Wait, does this function actually return anything?

wgranger: No, but a method marked `async` must return `Task` or `Task<T>` https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/async/#BKMK_ReturnTypesandParameters

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/CenterpieceViewModel.cs
wgranger: Good catch. I was a bit unsure on this, but I think the test code is worth another pass, considering.

wgranger:
I'd adjusted some of the tests here to keep all necessary items public or private. Many of the methods were able to easily convert back to `private` by using the `ICommand` attached to each method.

In a couple places, I've kept the method `public` when the method is triggered by another view model. Otherwise, I've kept the method private and have removed testing, as I've found many suggestions not to test private functions that aren't on the public API.

Unfortunately, Travis may not be a viable solution. Travis uses Mono for .NET build functionality, which doesn't appear to support building WPF applications. I'll have to do some research to see if there is a Windows-based CI capable of building WPF.

noahmalmed:
Regardless on what we do for CI. I think this is good to merge

PR #49, opened 2019-03-01
Add UI for User Rings over Subjects

wgranger:
Just as the title states, this PR adds rings to subjects users have interacted with. 

Solid rings reflect those users are currently classifying, low opacity rings w/o avatars represent those users have already classified on. 

A design change reflected on this PR. The inner yellow/green ring is _always_ present.

https://projects.invisionapp.com/d/main#/console/15093515/333377931/preview

### DIFF
--- a/GalaxyZooTouchTable/GalaxyZooTouchTable/Models/GalaxyRing.cs
+++ b/GalaxyZooTouchTable/GalaxyZooTouchTable/Models/GalaxyRing.cs
@@ -1,8 +1,9 @@
-﻿using System.Windows.Media.Imaging;
+﻿using GalaxyZooTouchTable.ViewModels;
+using System.Windows.Media.Imaging;
 
 namespace GalaxyZooTouchTable.Models
 {
-    public class GalaxyRing
+    public class GalaxyRing : ViewModelBase
     {
         const int RING_WIDTH_STEP = 16;
         const int INITIAL_RING_WIDTH = 56;
@@ -16,9 +17,18 @@ namespace GalaxyZooTouchTable.Models
         public int AvatarY { get; set; }
         public string UserColor { get; set; }
         public int YPos { get; set; }
+        public TableUser User { get; private set; }
+
+        private bool _currentlyClassifying = true;
+        public bool CurrentlyClassifying
+        {
+            get => _currentlyClassifying;
+            set => SetProperty(ref _currentlyClassifying, value);
+        }
 
         public GalaxyRing(int index, TableUser user)
         {
+            User = user;
             UserColor = user.ThemeColor;
             UserAvatar = user.Avatar;
             GetProperties(index);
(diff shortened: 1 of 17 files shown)


GalaxyZooTouchTable/GalaxyZooTouchTable/Models/GalaxyRing.cs
```diff
@@ -0,0 +1,95 @@
...
+            set => SetProperty(ref _currentlyClassifying, value);
+        }
+
+        public GalaxyRing(int index = 0, TableUser user = null)
+        {
+            if (user != null)
+            {
+                User = user;
+                UserColor = user.ThemeColor;
+                UserAvatar = user.Avatar;
```
noahmalmed: I think you should either store all of `user`, or store all of the values of `user` that you need.

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/GalaxyRing.cs
```diff
@@ -0,0 +1,95 @@
...
+        public void SetProperties(int index)
+        {
+            Diameter = INITIAL_RING_WIDTH + (RING_WIDTH_STEP * index);
+            AvatarX = INITIAL_AVATAR_RADIUS_POSITION + (index * (AVATAR_WIDTH_STEP));
+            AvatarY = INITIAL_AVATAR_RADIUS_POSITION + (index * (AVATAR_WIDTH_STEP));
+
+            AvatarPlacement();
+        }
+
+        private void AvatarPlacement()
```
noahmalmed: Should this maybe be called `PlaceAvater`?

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/SpaceViewModel.cs
```diff
@@ -30,6 +31,34 @@ public string SpaceCutoutUrl
...
+                            break;
+                        case RingNotifierStatus.IsHelping:
+                            SpaceViewGalaxy.RemoveRing(RingNotifier.User);
+                            break;
+                        default:
+                            break;
+                    }
+                } else if (RingNotifier.Status == RingNotifierStatus.IsLeaving)
+                {
+                    SpaceViewGalaxy.RemoveRing(RingNotifier.User);
```
noahmalmed: This logic seems a little weird. Why is `IsLeaving` not checked for in the switch?

wgranger: Because of the check for `Ring.Notifier.SubjectId == SpaceViewGalaxy.Subject.Id`, which checks that the current subject a user is classifying matches the correct subject on the board.

Instead, when a user is leaving the table, _all_ of their rings should be removed, including those subjects they have already seen and classified.

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/GalaxyRing.cs
wgranger: Yes, that would be a more active description.

wgranger:
This PR should be good to go. There are a couple bugs I've noticed in the build, but I don't think they're related to code changes here, so I'll open a new PR for them.

noahmalmed:
everything looks good. merging

PR #50, opened 2019-03-05
Fix Two Bugs

wgranger:
This PR fixes a couple bugs:

- There was a bad bug with the timer, mainly due to my misunderstanding of how `DispatcherTimer` works. You could easily recreate this bug if you opened several windows and toggled the panels to "reset" the timer (however, that reset wasn't working correctly). The timer now also resets whenever a touch is made on the classifier.

- Previously, when you dragged a galaxy over the an unused classifier/start button, the galaxy image would persist. This was due to an error in the `TapBehavior` where underlying elements weren't receiving events.

### DIFF
--- a/GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/StillThereViewModel.cs
+++ b/GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/StillThereViewModel.cs
@@ -9,8 +9,8 @@ namespace GalaxyZooTouchTable.ViewModels
 {
     public class StillThereViewModel : ViewModelBase
     {
-        public DispatcherTimer SecondTimer { get; set; }
-        public DispatcherTimer ThirtySecondTimer { get; set; }
+        public DispatcherTimer SecondTimer { get; set; } = new DispatcherTimer();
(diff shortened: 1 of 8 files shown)


GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/StillThereViewModel.cs
```diff
@@ -9,8 +9,8 @@ namespace GalaxyZooTouchTable.ViewModels
 {
     public class StillThereViewModel : ViewModelBase
     {
-        public DispatcherTimer SecondTimer { get; set; }
-        public DispatcherTimer ThirtySecondTimer { get; set; }
+        public DispatcherTimer SecondTimer { get; set; } = new DispatcherTimer();
+        public DispatcherTimer ThirtySecondTimer { get; set; } = new DispatcherTimer();
```
noahmalmed: Not sure if it's worth changing, but it may make more sense to have a counter that counts down from 30 on the `SecondTimer` instead of using the `ThirtySecondTimer`. May make the code more simple if you only had one timer in the code

wgranger: Good call. This will clean up the code quite a bit.

PR #53, opened 2019-03-12
Introduce Local Staging DB

wgranger:
Describe your changes.
This PR introduces a local database into the mix for retrieving subjects. This is the first step towards having an offline-first app. Notice, this branch contains functionality for a staging DB. The production DB will be added in later. Refer to the README on how to set up the database.

There's a good bit of refactoring in this branch as well. A good bit of it seeks to remove the ambiguity between `Subject` (straight from the Panoptes API) and `TableSubject` (optimized for the table). Having two similar items is confusing, so a good bit of the code here removes the `Subject` references.

# Review Checklist

- [X] Are tests passing?
- [X] Is the build free of output errors?

### DIFF
--- a/GalaxyZooTouchTable/GalaxyZooTouchTable/Models/TableSubject.cs
+++ b/GalaxyZooTouchTable/GalaxyZooTouchTable/Models/TableSubject.cs
@@ -38,20 +38,20 @@ namespace GalaxyZooTouchTable.Models
             RightAscension = ra;
             Declination = dec;
             SubjectLocation = location;
-            XYConvert(257.9, 23.23);
+            XYConvert();
         }
 
-        private void XYConvert(double CenterRightAscension, double CenterDeclination)
+        private void XYConvert()
         {
             int CutoutWidth = 1248; 
             int CutoutHeight = 432;
             const int ArcDegreeInSeconds = 3600;
 
             double DecRange = CutoutHeight * PlateScale / ArcDegreeInSeconds;
-            double RaRange = System.Math.Abs(CutoutWidth * PlateScale / ArcDegreeInSeconds / System.Math.Cos(CenterDeclination));
+            double RaRange = System.Math.Abs(CutoutWidth * PlateScale / ArcDegreeInSeconds / System.Math.Cos(SpaceNavigation.DEC));
 
-            double StartY = ((CenterDeclination - Declination) / DecRange * CutoutHeight) + (CutoutHeight / 2);
-            double StartX = ((CenterRightAscension - RightAscension) / (RaRange + Offset) * CutoutWidth) + (CutoutWidth / 2);
+            double StartY = ((SpaceNavigation.DEC - Declination) / DecRange * CutoutHeight) + (CutoutHeight / 2);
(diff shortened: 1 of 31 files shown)


GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/SpaceViewModel.cs
```diff
@@ -56,36 +82,124 @@ private void OnGalaxyInteraction(ClassificationRingNotifier RingNotifier)
...
+            }
+        }
+
+        private void GetSpaceCutout()
+        {
+            double WidenedPlateScale = 1.8;
+            SpaceCutoutUrl = $"http://skyserver.sdss.org/dr14/SkyServerWS/ImgCutout/getjpeg?ra={SpaceNavigation.RA}&dec={SpaceNavigation.DEC}&width=1248&height=432&scale={WidenedPlateScale}";
+        }
+
+        private void PrepareForNewPosition()
```
noahmalmed: For some reason this function name seems odd to me. I'm not sure I have a suggestion for a better name, but that's because I think I don't have a good understanding of what the function does.

noahmalmed: In general, I think it would improve clarity to make these functions less stateful. That is, instead of them mutating state, having the function take a `SpaceNavigation` and return a `Galaxy`. That way it's easier to track when state is actually changing.

noahmalmed: Or at least change the name of the function to something like
 `UpdateCurrentGalaxyWithNavigation (SpaceNavigation Navigation)`

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/TableSubject.cs
```diff
@@ -10,36 +10,35 @@ public class TableSubject
...
-        private readonly double PlateScale = 1.5;
-        private readonly double Offset = 0.03;
+        private readonly double WidenedPlateScale = 1.8;
         public Subject Subject { get; set; }
         public ObservableCollection<GalaxyRing> GalaxyRings { get; set; } = new ObservableCollection<GalaxyRing>();
+        public string Location { get; set; }
+        public string Id { get; set; }
 
-        public TableSubject(Subject subject, double TableRA = 0, double TableDEC = 0)
+        public TableSubject(string id, string location, double ra, double dec)
```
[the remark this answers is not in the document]
wgranger: No, they should be lower case.

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/SpaceViewModel.cs
wgranger: This might require some adjustment. `SpaceNavigation` is currently a global constant, but that might be a code smell. I'll make this change as the last commit on the changes I add, and we can decide from there if the change helps strengthen the code.

PR #60, opened 2019-04-09
Adjust drop area

### LINKED ISSUE #48: Keep User from Classifying Twice
With the introduction of Phase II, a user should not be allowed to drag a subject into their work space and classify more than once. A classifier should keep track of all subjects classified, and show a message to the user that more than one classifications on the same subject (per session) are not allowed.

### LINKED ISSUE #56: Drag Galaxy to Advance
Based on user input, dragging a galaxy into a classifier should begin a new classification. This should happen in several situations:

1. No galaxy is present in the classifier (current functionality)
2. Galaxy should _replace_ current galaxy shown
3. User is viewing classification results (drag and drop would advance to classifier)

Also, users expect dropping a galaxy anywhere on the classifier will begin classification for that galaxy (as opposed to only dropping in the highlighter activated "drop area").

wgranger:
Describe your changes.

This PR fixes several items:
- Allow users to advance the classifier and classify a galaxy when dropping a galaxy anywhere on the `SubjectViewer` or `ClassificationSummary`.
- Restricts users from classifying the same galaxy more than once by dragging it onto their classifier. 
- Fixed a bug where users could drop a galaxy onto their leveler or examples panel without the galaxy disappearing. This arose from the galaxy drop event being handled before the drag overlay could clean up and remove the galaxy.

Fixes #48 
Closes #56 

# Review Checklist

- [x] Are tests passing?

### DIFF
--- a/GalaxyZooTouchTable/GalaxyZooTouchTable/Lib/GalaxyAdorner.cs
+++ b/GalaxyZooTouchTable/GalaxyZooTouchTable/Lib/GalaxyAdorner.cs
@@ -7,39 +7,39 @@ using System.Windows.Media;
 namespace GalaxyZooTouchTable.Lib
 {
     public class GalaxyAdorner : Adorner
+    {
+        public Point Location;
+        public Point Offset;
+        public EventHandler<TouchEventArgs> UnsubscribeEvent { get; set; }
+
+        public GalaxyAdorner(FrameworkElement adornedElement, Point offset)
+          : base(adornedElement)
         {
-            public Point Location;
-            public Point Offset;
-            public EventHandler<TouchEventArgs> UnsubscribeEvent { get; set; }
-
-            public GalaxyAdorner(FrameworkElement adornedElement, Point offset)
-              : base(adornedElement)
-            {
-                Offset = offset;
-                DataContext = adornedElement.DataContext;
-                IsHitTestVisible = false;
-            }
-
-            public void UpdatePosition(Point location)
-            {
-                Location = location;
-                this.InvalidateVisual();
-            }
-
-            protected override void OnRender(DrawingContext drawingContext)
-            {
-                var adornerLocation = Location;
-                adornerLocation.Offset(-Offset.X, -Offset.Y);
-
-                Rect adornedElementRect = new Rect(adornerLocation, AdornedElement.DesiredSize);
-
-                var visualBrush = new VisualBrush(AdornedElement);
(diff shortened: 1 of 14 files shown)


shaunanoordin:
Update: I just realised that I can't run this project on my work machine - _"This project type is not supported by Visual Studio Community 2019 for Mac"_ - so I'll need to look into setting up one of my Windows machines if I want to be helpful.

wgranger:
Interesting. I'm running an older version of Visual Studio, so I'll have to look into what's happening there.

wgranger:
I doubt the Mac version is capable of running the .exe that this compiles to.

chelseatroy:
My comments:

Currently testing all the functionality described in the fix list above. It does look like the galaxy circle needs to be _totally_ on the workspace (no partial) or else it doesn't "stick." I think that's fine, but I leave it up to you if you would like that to work differently.

GalaxyZooTouchTable/GalaxyZooTouchTable/Lib/GalaxyAdorner.cs
```diff
@@ -7,39 +7,39 @@
...
+        public EventHandler<TouchEventArgs> UnsubscribeEvent { get; set; }
 
-            public GalaxyAdorner(FrameworkElement adornedElement, Point offset)
-              : base(adornedElement)
-            {
-                Offset = offset;
-                DataContext = adornedElement.DataContext;
-                IsHitTestVisible = false;
-            }
+        public GalaxyAdorner(FrameworkElement adornedElement, Point offset)
```
chelseatroy: i like this name :)

GalaxyZooTouchTable/GalaxyZooTouchTable/Views/SubjectViewer.xaml.cs
```diff
@@ -1,15 +1,37 @@
...
     {
         public SubjectViewer()
         {
             InitializeComponent();
         }
+
+        public bool IsUnder(Point p)
+        {
+            throw new System.NotImplementedException();
+        }
```
chelseatroy: I suspect you're forced to implement this. Is that accurate?

wgranger: Yes, this is required because `SubjectViewer` inherits from the `IDroppableInterface`, even though the `IsUnder` method isn't used here. I may adjust this in the future, but it's worth keeping around as a reminder if I plan to restructure this later to truly implement the `IsUnder` function.

Issue #52, opened 2019-03-12
Offline First


wgranger:
The table should be prepared to operate without internet. Several items must be changed to accomplish this:

- [x] If possible, use static workflow values instead of retrieving resource from Panoptes
- [x] Use a classification queue to keep failed classifications saved locally
- [x] Keep track of classification counts on the local database
- [x] Use classification counts in place of calling for Caesar reductions

wgranger:
What should happen when the app first loads without internet connection? Currently, the app will fail when the workflow resource resolves as `null`. It seems important to have internet for this initial workflow call, as the resource is used for workflow versioning on submitted classifications and finding workflow tasks.

My initial thought is to have an initial warning on app load that would say, "please connect to the internet to start the app," or something similar.

wgranger:
Now that the first phase of this is complete, the final step is:

- [ ] Use locally stored subjects to load images instead of relying on internet to fetch source

PR #62, opened 2019-04-29
Offline first

wgranger:
Describe your changes.
This PR allows the touch table to work offline. This will be marked WIP until https://github.com/zooniverse/panoptes-net-client/pull/9 is merged on the .NET client. Afterwards, the client will have to be updated on Nuget and the dependency updated here. 

Working towards #52 

# Review Checklist

- [x] Are tests passing?

### DIFF
--- a/GalaxyZooTouchTable/GalaxyZooTouchTable.Tests/Data/OfflineWorkflow.xml
+++ b/GalaxyZooTouchTable/GalaxyZooTouchTable.Tests/Data/OfflineWorkflow.xml
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8" ?>
 <root>
   <Workflow name="GalaxyZoo">
-    <version>0.0</version>
+    <version>0.4</version>
     <firstTask>T0</firstTask>
     <tasks>
       <task name="FirstTask">
--- a/GalaxyZooTouchTable/GalaxyZooTouchTable.Tests/Data/OfflineWorkflow.xml
+++ b/GalaxyZooTouchTable/GalaxyZooTouchTable.Tests/Data/OfflineWorkflow.xml
@@ -1,0 +1,14 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<root>
+  <Workflow name="GalaxyZoo">
+    <version>0.0</version>
+    <firstTask>T0</firstTask>
+    <tasks>
+      <task name="FirstTask">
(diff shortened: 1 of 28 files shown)


wgranger:
## Update: May 23

This PR is now ready to review. This stayed WIP a bit longer than I wanted because the `ClassificationPanelViewModel` was being refactored to improve structure and remove unnecessary items. There's some more refactoring I'd like to do there, but I'd like some other PRs merged before making any more changes.

The biggest change to offline-first in the most recent commits improved how we handle GraphQL queries from Caesar. 

**Originally:** Each classifier (6, around the table) would load a subject queue (10 subjects) on app load. Caesar would then be queried for each new "CurrentSubject" at each classifier. The local DB would not be updated for each answer (smooth, features, star).

**Changes:** Now, classifiers do not begin with a subject queue. One is fetched only if a user chooses "random galaxy" after the first classification.

Each time a new space cutout is loaded in the center of the app (including the initial app load) Caesar is queried for each galaxy. If Caesar returns classification results, we can assume those results are accurate and update our local DB accordingly. 

If Caesar returns nothing, we assume there is no information or Caesar is down. Either way, the local DB is not updated.

On classification, we find a subject's classification counts from the local DB, increment the total count and answer selection, and update those results to the local DB.

**tl:dr**
- We use Caesar for reliable classification counts and local DB updating
- If Caesar is down, we make no update on subject load
- The local DB is updated with each classification

wgranger:
Still WIP until the [.NET client PR](https://github.com/zooniverse/panoptes-net-client/pull/9) is merged

wgranger:
I've updated the .NET client on Nuget, updated the dependencies here, and did a couple last minute checks. Should be good for a review.

wgranger:
Rebased and ready for review.

chelseatroy:
Hi Will!

Happy to review this. What steps would you recommend I follow to make sure it works? At what point should I turn off wifi, et cetera?

Thanks!
Chelsea

wgranger:
I'm glad you're asking this. You can turn off wifi at any point, and the app should continue to work. The only issue is that subject images will contain a placeholder (the U!Scientist logo) there is no internet connection to download the image. However, when internet comes back, the images should download.

chelseatroy:
Will look at this more again next week! The main thing, as you and I saw together, is that on first pass, I am not seeing a placeholder image for each of the galaxies. I'll look at this again on Monday, though, maybe it's some weird thing. Might be worth looking into though.

wgranger:
Good call. I decided to remove the fallback image after a conversation in Slack. A fallback image isn't the best solution, because a user wouldn't be able to classify the image with a fallback icon. Instead, I'll push up a separate PR in the future that addresses this. 

The likely solution is that we will have the subject images accessible locally, so users are still able to see images even if the internet is down.

PR #63, opened 2019-05-09
Change Icons and Ring Styles

### LINKED ISSUE #59: Ring Styles Matching User Border Styles
Currently, each ring around a user's galaxy is solid and round. However, each ring should instead match the dot pattern of each user's border. The border should match that of each users [start button](https://projects.invisionapp.com/d/main#/console/15093515/314161096/preview).

wgranger:
Describe your changes.
This PR changes user icons to the more abstract shapes we'll be using on the final version. Also, each username has been changed throughout the code base to the color shown (GreenUser, BlueUser, etc.). Ring styles, when classifying a galaxy, have also been changed from a solid circle to the dash style of a particular user.

Fixes #59  .

Design template [here](https://projects.invisionapp.com/share/WFNHU46KV6J#/screens/354519105_10-1_All_Open_New_Icons_1)

# Review Checklist

- [x] Are tests passing?


GalaxyZooTouchTable/GalaxyZooTouchTable/Models/GalaxyRing.cs
```diff
@@ -68,21 +70,21 @@ private void PlaceAvatar()
...
+                    case "AquaUser":
                         AvatarY = 0;
                         AvatarX = Diameter / 2;
                         break;
-                    case "PersonUser":
+                    case "PurpleUser":
                         AvatarX *= -1;
                         break;
-                    case "EarthUser":
+                    case "GreenUser":
```
chelseatroy: We have a relatively large case statement--which is OK, since it's pretty clear what it does.

Another option might be to put attributes for `XPosition` and `YPosition` onto the `user` object that we're passing into `GalaxyRing`,  and then assign `AvatarX` and `AvatarY` to those attributes inside of `PlaceAvatar()`.

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/TableUserFactory.cs
```diff
@@ -36,51 +36,51 @@ public bool Active
...
-    public class PersonUser : TableUser
+    public class PurpleUser : TableUser
     {
-        public override string Name => "PersonUser";
+        public override string Name => "PurpleUser";
         public override string ThemeColor => "#A5A2FB";
-        public override BitmapImage StartButton => new BitmapImage(new Uri("../Images/StartIcon/Person.png", UriKind.Relative));
-        public override BitmapImage Avatar => new BitmapImage(new Uri("../Images/Avatars/Person.png", UriKind.Relative));
+        public override BitmapImage Avatar => new BitmapImage(new Uri("../Images/Avatars/purple.png", UriKind.Relative));
+        public override string DashArray => "0.25 1.5 2 1.5";
```
chelseatroy: As per the comment above, in here is where I might put those `XPosition` and `YPosition` attributes.

GalaxyZooTouchTable/GalaxyZooTouchTable/Models/GalaxyRing.cs
wgranger: One issue with that is that a ring's `AvatarX` and `AvatarY` will change values whenever a user leaves the table and their ring around a subject disappears. Thus, the X and Y positions are changing depending on how many users have classified a subject and in which order (index value) the user's ring appears. 

However, most of this display stuff probably shouldn't be done in the Models, so I think your point is valid that there is probably a better solution to be had. I'll timebox this and see if I can find an alternative.

wgranger:
Alright, I think I was able to do something similar to what you suggested. The avatar is now positioned base on where a workstation appears on the table (horizontally: left or right, and vertically: top, middle, or bottom. This adds a few extra lines of code, but it removes some of the dependencies on user names. 

As stated earlier, I'd eventually like to calculate view logic somewhere besides the model, but that is a larger structural change that will require a separate PR.

PR #66, opened 2019-05-15
Fix Design Bugs

wgranger:
Describe your changes.
Several design bugs are addressed here:
- Fixes bug where leveler wasn't working
- Updates notification bell to match appearance of leveler (when closed)
- Keeps "drop galaxy here" box from turning yellow on tap
- Introduces PushButton component to allow buttons to keep downstate styles on multitouch
- Adds new galaxy example images to the side pullout

# Review Checklist

- [x] Are tests passing?


GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ClassificationPanelViewModel.cs
```diff
@@ -54,7 +54,13 @@ public TableSubject CurrentSubject
 
         public NotificationsViewModel Notifications { get; private set; }
         public ExamplesPanelViewModel ExamplesViewModel { get; private set; }
-        public LevelerViewModel LevelerViewModel { get; private set; }
+
+        private LevelerViewModel _levelerViewModel;
+        public LevelerViewModel LevelerViewModel
+        {
+            get => _levelerViewModel;
+            set => SetProperty(ref _levelerViewModel, value);
```
chelseatroy: What's the relationship between the private and the public LevelerViewModels?

wgranger: This is a common setup for the MVVM pattern to update a property through the 'INotifyPropertyChanged' interface, which the `ViewModelBase` inherits from. This is necessary so the `ClassificationPanel` view using this view model will get an updated `LevelerViewModel` whenever the property is changed, which happens on this line https://github.com/zooniverse/galaxy-zoo-touch-table/blob/master/GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ClassificationPanelViewModel.cs#L239 to reset the leveler counts.

wgranger:
Made a few small changes, mainly naming of `PushButton` to `MultiTouchButton`.

PR #68, opened 2019-06-13
Create Classification Summary VM

### LINKED ISSUE #58: Different messages when classifying
Currently, a user will always see "you just science'd" after making a classification. We discussed showing a variety of different messages to surprise users and have them continue classifying.

wgranger:
Describe your changes.
This PR shows a different random message after a user classifies a subject.

**June 18 Update**: Making changes here to abstract out a classification summary view model which will also fix errors with stats appearing after a classification.

Fixes #58 .

# Review Checklist

- [x] Are tests passing?


wgranger:
Marking this WIP. I have to do some work on top of this branch, and rather than leading to a messy rebase, I'm going to add some commits to this.

wgranger:
This is now ready for review. My most recent commits were done to abstract a new `ClassificationSummaryViewModel` from the `ClassificationPanelViewModel`. Basically, the `ClassificationPanelViewModel` is a bit too bloated and shouldn't handle functionality of other views. Classification counts should now be correct on the summary screen (they were originally broken after the offline first merge).

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ClassificationSummaryViewModel.cs
```diff
@@ -0,0 +1,69 @@
...
+        public event Action<TableSubject> DropSubjectDelegate = delegate { };
+
+        readonly List<string> SummaryStrings = new List<string>
+        {
+            "Nice work! You just science'd.",
+            "Thanks for your help!",
+            "You're a real scientist!",
+            "Don't stop now!",
+            "You can do this at home on zooniverse.org."
+        };
```
chelseatroy: Ohhhh I see, we randomly select between these for a message upon a classification happening.

chelseatroy:
Love it. Two notes:

1. When we do random selection, it is theoretically possible (though unlikely to happen often) for the same message to be shown a whole bunch of times in a row. I don't actually think that's a problem here, but it's worth noting that that's a possibility. The fix would be to deliberately iterate through the messages. This solution is certainly enough overhead that I _don't_ think it would be worthwhile.

2. "You can do this at home on zooniverse.org" shows up pretty small relative to the other messages. Options:
 - Leave it alone. It's certainly still legible. I'd defer to a designer on how "arresting" the message needs to be.
 - Allow the message to spill onto multiple lines. At our current font size this would put "zooniverse.org" on the next line. I would hope that wouldn't confuse people, but I don't know for sure. Again would defer to a designer.
 - Shorten the message to "Keep going at zooniverse.org." Unfortunately even something like "Do this at home on zooniverse.org" still results in a smaller font.

I'm not partial to any of these solutions, and I think it's probably fine the way it is. Just wanted to mention what I had observed and offer some possible alternatives if the observation merits attention.

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ClassificationSummaryViewModel.cs
wgranger: Yeah, in an earlier commit this was done in the `App.xaml` file, but I moved it to the class using the string so tests would pass. I think it makes more sense at the class-level as well.

wgranger:
1. Yes, I think keeping the string randomized is a suitable solution. 
2. Good call. I resized this to fit because the string was much longer than other strings. Let me check with Becky this resizing solution is suitable or if we should try another solution.

wgranger:
The url has actually changed for that summary string! I've made the change and changed the text a bit to keep it on the short side. It still resizes a bit, but I've confirmed with Becky that this is acceptable.

chelseatroy:
Lookin great!

Issue #70, opened 2019-06-20
June 2018 Design Updates


wgranger:
A few design updates based on testing the table upstairs un early June:

- [x] Start button should reach edge of table
- [x] Users are double tapping the "submit" button and skipping the summary screen
- [x] Galaxies should "wiggle" to show they can be dragged
- [x] Fix classification counts on the summary page
- [x] Answer buttons should be disabled until a galaxy is visible in the classifier
- [x] "Help classify the galaxy below" is showing when there isn't a galaxy present
- [x] Examples panel should remain open at all times (shouldn't collapse)
- [x] Expand touch area to drag galaxies
- [x] Wait 10 seconds before galaxies wiggle
- [x] Change h1 to ‘Help us classify these galaxies’
- [x] Examples panel current `<` should appear as a back button
- [x] Add an illustration of a finger dragging a galaxy instead of the phrase ‘drag a galaxy’

PR #72, opened 2019-06-27
June Design Updates

wgranger:
Describe your changes.
Working towards #70, but missing the wiggling galaxies, as I imagine that will be a bigger fix.

# Review Checklist

- [x] Are tests passing?

### DIFF
--- a/GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ClassificationSummaryViewModel.cs
+++ b/GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ClassificationSummaryViewModel.cs
@@ -2,6 +2,7 @@
 using GalaxyZooTouchTable.Utility;
 using System;
 using System.Collections.Generic;
+using System.Threading.Tasks;
 using System.Windows.Input;
 
 namespace GalaxyZooTouchTable.ViewModels
@@ -10,6 +11,7 @@ namespace GalaxyZooTouchTable.ViewModels
     {
         static Random random = new Random();
         public NotificationsViewModel Notifications { get; private set; }
+        bool ButtonsDisabled = false;
 
         public ICommand RandomGalaxy { get; private set; }
         public ICommand ChooseAnotherGalaxy { get; private set; }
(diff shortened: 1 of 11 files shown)


chelseatroy:
So, I'm systematically going through these, and the changes as described appear to work :).

I did notice one thing, and this might be for a different PR: When I open a panel for each of two users and have one ask another for help with a galaxy, the "helping" one's pattern ring does not appear around the galaxy. I don't know if that's the intended design, but it seemed like a possible inconsistency, so I thought I'd mention it. I don't think it's a show-stopper.

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ClassificationSummaryViewModel.cs
```diff
@@ -43,12 +45,25 @@ public void ProcessNewClassification(string subjectLocation, ClassificationCount
...
         {
-            RandomGalaxy = new CustomCommand(OnRandomGalaxy);
-            ChooseAnotherGalaxy = new CustomCommand(OnChooseAnotherGalaxy);
+            RandomGalaxy = new CustomCommand(OnRandomGalaxy, CanBeginNewClassification);
+            ChooseAnotherGalaxy = new CustomCommand(OnChooseAnotherGalaxy, CanBeginNewClassification);
+        }
+
+        private bool CanBeginNewClassification(object obj)
+        {
+            return !ButtonsDisabled;
```
chelseatroy: I'm a big fan of using clear naming the way you have done here rather than sticking the inverse of ButtonsDisabled directly into the CustomCommand constructor 🥇

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ExamplesPanelViewModel.cs
```diff
@@ -76,20 +63,5 @@ public bool CanSelectItem(object sender)
...
-
-        public bool CanOpen(object sender)
-        {
-            return !IsOpen;
-        }
-
-        public bool CanToggle(object sender)
-        {
-            return true;
-        }
```
chelseatroy: Perfect. Since it's now never closed we don't need any of this

wgranger:
Hmm. That's interesting. I'm unable to replicate that behavior. You say User Red asks User Blue for help. User Blue accepts but there is no Blue ring around the galaxy? I'm seeing the Blue ring pop up, unless there are rare occasions when it doesn't?

PR #74, opened 2019-06-28
Use secondary cutout

### LINKED ISSUE #54: What to do when cutout service is down
We currently have a dependency on the Sloan cutout service. However, when the service is down, out space cutout will currently be blank.

We should use the dark energy survey as the backup service, but I'm still not sure a sufficient solution in the case both services are down.

wgranger:
Describe your changes.
With the changes here, we now rely on two cutout services (DECALS & Sloan) in the instance one of those services goes down. If both are down, or we lose internet connection, the background space image will be blank.

A few things of note:
- When a cutout service isn't responding, we take a ten minute break from trying to fetch images there.
- Unfortunately DECALS only returns squared images, so I have to stitch together three images for the background.
- DECALS takes about 10-15 seconds to load while navigating while Sloan takes around one second. Unfortunately, our subjects come from the DECALS survey.

Closes #54 

# Review Checklist

- [x] Are tests passing?

### DIFF
--- a/GalaxyZooTouchTable/GalaxyZooTouchTable.Tests/ViewModels/SpaceViewModelTests.cs
+++ b/GalaxyZooTouchTable/GalaxyZooTouchTable.Tests/ViewModels/SpaceViewModelTests.cs
@@ -12,6 +12,7 @@ namespace GalaxyZooTouchTable.Tests.ViewModels
     {
         private SpaceViewModel _viewModel { get; set; }
         private Mock<ILocalDBService> _localDBServiceMock = new Mock<ILocalDBService>();
+        private Mock<ICutoutService> _cutoutServiceMock = new Mock<ICutoutService>();
 
         public SpaceViewModelTests()
         {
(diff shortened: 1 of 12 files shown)


chelseatroy:
I'm not sure how to knock on one service and then the other to check this without manually doing something to the connection in the code itself, but I can confirm that whatever service loads as the default looks good, and I can also confirm that the view looks the way it's supposed to (based on the above description) with zero internet connection.

@wgranger, how were you testing this with each service?

wgranger:
So... I've done a few things, but they were reliant on me changing the code and commenting out some items. I'd like to get some tests in here to cover certain conditions, but I'm a bit under time constraints to get those tests in now. You could change the URL for the cutout service to something you know will timeout (there are sample urls online to replicate this).

wgranger:
The most recent commits move away from the method of stitching together three images for a rectangular DECALS cutout (see first note at top). Although the method worked, it was taking too long (sometimes 30 seconds) to stitch together three images for each location around the table.

Instead, the table now retrieves three different DECAL images from the cutout service and displays them aside one another on the `SpaceView.xaml` view. This saves a ton of time and frustration for the user.

PR #75, opened 2019-06-28
Use Local subjects

### LINKED ISSUE #52: Offline First

wgranger:
Describe your changes.
Just a couple changes needed here to have the app use local subjects (if they exist). Local subjects should be put in the system's Documents folder. I have the local subjects on a removable hard drive, and they account for about 100GB of space. This should be the final step needed for offline-first.

See more info in the updated README here.

Closes #52 

# Review Checklist

- [x] Are tests passing?

### DIFF
--- a/GalaxyZooTouchTable/GalaxyZooTouchTable.Tests/Mock/PanoptesServiceMockData.cs
+++ b/GalaxyZooTouchTable/GalaxyZooTouchTable.Tests/Mock/PanoptesServiceMockData.cs
@@ -19,7 +19,7 @@ namespace GalaxyZooTouchTable.Tests.Mock
 
         public static TableSubject TableSubject()
         {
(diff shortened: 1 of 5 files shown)


chelseatroy:
Looks good! I recommend including in the README how this works (i.e., that it queues the classifications locally and sends them all when the internet returns, or loses them all if the app is closed).

wgranger:
I've added a quick blurb on offline first in the README. Let me know if you think it's missing anything.

PR #76, opened 2019-07-09
Additional changes

### LINKED ISSUE #73: Prepare Production App
During development, the app has relied on a staging project for everything. However, we'll have to make sure the production workflow works as expected. This shouldn't require many (if any) code changes, but is worth checking before the app goes on the floor.

wgranger:
Describe your changes.

The app should be production ready after this PR merges. Added a couple items based on a final design review:
- Update workflow version for static workflow
- Add hand icon to drop zone
- Allow random subject to load on tapping drop zone
- Switch "random subject" and "choose subject" background button colors
- Handles a bug where the deploy can't find offline workflow info
- Updates QR code to point to U!Scientist
- Change all adler.uscientist.org links to uscientist.org

Closes #73 

# Review Checklist

- [x] Are tests passing?


wgranger:
WIP as there are a few issues with the deploy that might require some additional fixes here.

wgranger:
This should be good to go. The most recent commit fixes the following:

When deploying, the app has trouble finding the `.xml` file that the OfflineWorkflow was built off of, and a couple errors occur. Because of this, I've just created a static workflow instance globally that the app can refer to. Errors are gone, and the app should build to the table now.

PR #81, opened 2019-07-11
Account for Retired Galaxies

### LINKED ISSUE #77: What Happens When Subjects Retire?
The retirement rate is currently at 25. However, a subject can be classified more than 25 times, but this will look awkward with the current UI (eg. 27/25 classifications?).

What should we do when a subject gets to 25 classifications (retires)? Still allow users to interact with the galaxies but disable classification? Dim out and disable to galaxies?

### LINKED ISSUE #78: Occasionally Move Map
I can see a situation on the table where users will not spot the "move map" button and be stuck on a single location for a while. This could be a problem if all galaxies on that spot get classified and users don't know they should move the map. 

Because of this, there needs to be a check frequently that there are still galaxies to be classified at the current table location.

wgranger:
Describe your changes.
This PR brings in a modal that appears if a subject is retired (has hit 25 classifications). This _only_ occurs when dropping a subject into a classifier for a couple reasons.

- A user should always receive subjects with the lowest classification count when asking for a random subject. In the unlikely case all subjects are retired, I don't want to run into an infinite loop where no random subject can be retrieved because they're all retired and the modal is constantly shown since no subject can be retrieved.
- I think users should still be able to ask others at the table for advice on a subject, even if the subject is moving past retirement. There is no harm in a subject submitting more than 25 classifications to Panoptes, and users should still be able to interact with one another. 

Unrelated, I also removed the `IsOpen` variable from several `ExamplePanel` views, as that variable was no longer being used by the view model, or anything in particular.

Fixes #77 
Fixes #78 

# Review Checklist

- [x] Are tests passing?

### DIFF
--- a/GalaxyZooTouchTable/GalaxyZooTouchTable.Tests/ViewModels/ClassificationPanelViewModelTests.cs
+++ b/GalaxyZooTouchTable/GalaxyZooTouchTable.Tests/ViewModels/ClassificationPanelViewModelTests.cs
@@ -23,6 +23,7 @@ namespace GalaxyZooTouchTable.Tests.ViewModels
             _panoptesServiceMock.Setup(dp => dp.CreateClassificationAsync(It.IsAny<Classification>()))
                 .ReturnsAsync(new ClassificationCounts(1,0,0,1));
 
+            _localDBServiceMock.Setup(dp => dp.CheckIfSubjectRetired(It.IsAny<string>())).Returns(false);
             _localDBServiceMock.Setup(dp => dp.GetLocalSubject(It.IsAny<string>())).Returns(PanoptesServiceMockData.TableSubject());
             _localDBServiceMock.Setup(dp => dp.GetQueuedSubjects()).Returns(PanoptesServiceMockData.TableSubjects());
 
@@ -160,6 +161,34 @@ namespace GalaxyZooTouchTable.Tests.ViewModels
         {
             _viewModel.ShowCloseConfirmation.Execute(null);
             Assert.True(_viewModel.ShowOverlay);
+        }
+
+        [Fact]
+        private void ShouldNotShowSubjectRetired()
+        {
+            _viewModel.Load();
+            _viewModel.DropSubject(PanoptesServiceMockData.TableSubject());
+            Assert.False(_viewModel.ShowRetirementModal);
+        }
+    }
+
+    public class ClassificationPanelRetiredSubjectTests
+    {
+        private ClassificationPanelViewModel _viewModel { get; set; }
+        private Mock<IPanoptesService> _panoptesServiceMock = new Mock<IPanoptesService>();
+        private Mock<ILocalDBService> _localDBServiceMock = new Mock<ILocalDBService>();
+
+        public ClassificationPanelRetiredSubjectTests()
+        {
+            _localDBServiceMock.Setup(dp => dp.CheckIfSubjectRetired(It.IsAny<string>())).Returns(true);
+            _viewModel = new ClassificationPanelViewModel(_panoptesServiceMock.Object, _localDBServiceMock.Object, new BlueUser());
+        }
+
+        [Fact]
+        private void ShouldShowRetirementModal()
+        {
+            _viewModel.DropSubject(PanoptesServiceMockData.TableSubject());
+            Assert.True(_viewModel.ShowRetirementModal);
         }
     }
 }
--- a/GalaxyZooTouchTable/GalaxyZooTouchTable.Tests/ViewModels/ClassificationPanelViewModelTests.cs
+++ b/GalaxyZooTouchTable/GalaxyZooTouchTable.Tests/ViewModels/ClassificationPanelViewModelTests.cs
@@ -23,7 +23,6 @@ namespace GalaxyZooTouchTable.Tests.ViewModels
             _panoptesServiceMock.Setup(dp => dp.CreateClassificationAsync(It.IsAny<Classification>()))
                 .ReturnsAsync(new ClassificationCounts(1,0,0,1));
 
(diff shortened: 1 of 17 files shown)


GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ClassificationPanelViewModel.cs
```diff
@@ -356,8 +376,16 @@ public void DropSubject(TableSubject subject)
             GlobalData.GetInstance().Logger?.AddEntry("Drop_Galaxy", User.Name, subject.Id, CurrentView);
             if (CheckAlreadyCompleted(subject)) return;
             if (CurrentView == ClassifierViewEnum.SummaryView) CurrentView = ClassifierViewEnum.SubjectView;
-            LoadSubject(subject);
-            NotifySpaceView(RingNotifierStatus.IsCreating);
+            if (_localDBService.CheckIfSubjectRetired(subject.Id))
```
wgranger: If you change this line to `if (true)` in your local code then you should see the retired modal appear every time you drag a subject into the classifier. Unfortunately, the current subject set doesn't have many retired subjects hitting the 25 classification threshold.

wgranger:
Marking this WIP while I add several more items needed for retirement.

wgranger:
Good bit of commits going on here. I'll explain the main changes:
- Open retired galaxy modal if a galaxy is retired, restricting a user from classifying
- Mark a subject `isRetired` if it contains more than 25 classifications
- Fixes some issues with moving the map north, south, east, and west. 
- On closing the map, move to a new area of space if the current subjects are all retired
- Pulse movemap buttons around the center space area every ten seconds
- Account for edge case of all galaxies retired by returning a random subject with the least amount of classifications.
- Dim the green ring around a subject/galaxy if that item is retired

GalaxyZooTouchTable/GalaxyZooTouchTable.Tests/ViewModels/ClassificationPanelViewModelTests.cs
```diff
@@ -161,5 +161,34 @@ private void ShouldShowOverlay()
             _viewModel.ShowCloseConfirmation.Execute(null);
             Assert.True(_viewModel.ShowOverlay);
         }
+
+        [Fact]
```
chelseatroy: The `[Fact]` indicating unit tests is my favorite thing about xUnit for C#

GalaxyZooTouchTable/GalaxyZooTouchTable/Services/ILocalDBService.cs
```diff
@@ -9,9 +9,9 @@ public interface ILocalDBService
     {
         TableSubject GetLocalSubject(string id);
 
-        List<TableSubject> GetLocalSubjects(SpaceNavigation currentLocation);
+        List<TableSubject> GetLocalSubjects(SpaceNavigation currentLocation, bool returnIfRetired = false);
```
chelseatroy: Is there a circumstance under which would we want returnIfRetired to be false, or do we always want this to be the case?

GalaxyZooTouchTable/GalaxyZooTouchTable/Services/LocalDBService.cs
```diff
@@ -108,7 +113,11 @@ public List<TableSubject> GetSubjects(string query, SpaceNavigation currentLocat
                     Messenger.Default.Send(ErrorMessage, "DatabaseError");
                 }
                 UpdateDBFromIds(idsToUpdate);
-                return Subjects;
+
+                if (returnIfRetired)
+                    return Subjects;
+                else
+                    return Subjects.Any(subject => !subject.IsRetired) ? Subjects : new List<TableSubject>();
```
chelseatroy: Ohhh, I see. OK ignore question above!

GalaxyZooTouchTable/GalaxyZooTouchTable/Services/ILocalDBService.cs
wgranger: There's a use case for both circumstances. I'm mainly trying to avoid being caught in an infinite loop where the local database is constantly being queried for subjects but can't find any because they're all retired. This is a definite possibility with the added [classification limits](https://github.com/zooniverse/galaxy-zoo-touch-table/pull/81/files/e204073f77317689fa3b461bae38be53b383beca#diff-7555746014045015b676c065b11e05dfR22) to each many of the database queries. 

What would happen if _every_ subject on the table already has 25 classifications? It's rare that would happen, but a possibility. I don't want the table to crash or stop navigating because there are no subjects to navigate. In this case, a subject should still be loaded, even if it has retired. 

Ex: We move west by running [this method](https://github.com/zooniverse/galaxy-zoo-touch-table/pull/81/files/e204073f77317689fa3b461bae38be53b383beca#diff-03461c6521d90ee0752bff2604107b44R202). The app first checks if any galaxies are one cutout image west. If [none exist there](https://github.com/zooniverse/galaxy-zoo-touch-table/pull/81/files/e204073f77317689fa3b461bae38be53b383beca#diff-03461c6521d90ee0752bff2604107b44R253), the app tries to find the [next western-most galaxy](https://github.com/zooniverse/galaxy-zoo-touch-table/pull/81/files/e204073f77317689fa3b461bae38be53b383beca#diff-7555746014045015b676c065b11e05dfR162). That process looks like this: 

- Find the galaxy with the next highest right ascension (RA) value, ascending
- If that doesn't exist, find the galaxy with the lowest RA value, as we're gone around in a circle in space where 360 degrees RA goes back to zero degrees
- If you still can't find a galaxy because they're all retired, just load any random galaxy with the least amount of classifications

The third point is where we want to return subjects from the local database, even if they're retired.

PR #84, opened 2019-07-15
Fix queue bug

wgranger:
Describe your changes.
After deploying the table to the floor, a bug cropped up where the app will crash if the queued classifications are modified in place. This can happen two ways:

- A user double taps "submit classification" before their classification has been sent, resulting in the classification queue being added to while it is also being checked for classifications in a `foreach` loop.
- Two users click "submit classification" at the same time, resulting in the `PanoptesService` singleton (one instance) trying to modify the classification queue while it is being checked for a pending classification. This was remedied by giving each classifier its own `PanoptesService` instance and, thus, its own classification queue.

I also removed the `IsOpen` tag that is unnecessarily being referenced in the `ExamplesPanel` and the faulty `User.Name` binding in the same view.

# Review Checklist

- [x] Are tests passing?

### DIFF
--- a/GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ClassificationPanelViewModel.cs
+++ b/GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ClassificationPanelViewModel.cs
@@ -19,6 +19,7 @@ namespace GalaxyZooTouchTable.ViewModels
         private ILocalDBService _localDBService;
         private List<CompletedClassification> CompletedClassifications { get; set; } = new List<CompletedClassification>();
         private DispatcherTimer StillThereTimer { get; set; } = new DispatcherTimer();
+        private bool IsSubmittingClassification;
 
         public Classification CurrentClassification { get; set; } = new Classification();
         public List<TableSubject> Subjects = new List<TableSubject>();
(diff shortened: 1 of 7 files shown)


wgranger:
I've added a few extra commits here to handle some bugs I was seeing today, mainly related to existential checks.

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/ClassificationPanelViewModel.cs
```diff
@@ -191,7 +192,12 @@ private void LoadCommands()
             OpenClassifier = new CustomCommand(OnOpenClassifier);
             SelectAnswer = new CustomCommand(OnSelectAnswer);
             ShowCloseConfirmation = new CustomCommand(OnShowCloseConfirmation);
-            SubmitClassification = new CustomCommand(OnSubmitClassification);
+            SubmitClassification = new CustomCommand(OnSubmitClassification, CanSubmitClassification);
```
chelseatroy: This is ultra minor and I don't think you have to change this, but it would make sense for the custom command to take parameters in the order it will use them (check if can submit classification, then act upon the submission). That would mean switching the order of the constructor params here.

wgranger: This does make sense. Unfortunately, I'd have to change the order of the arguments for all custom commands on the app, which is around 100 or so. This might be something worth doing a little further down the line when some of the current design issues are closed.

PR #86, opened 2019-07-17
Provide More Existential Checks

wgranger:
Describe your changes.
There's still a bug popping up with the touchLeave method

Fixes # .

# Review Checklist

- [x] Are tests passing?


wgranger:
If this code fails to run, a user would not be able to drag a galaxy into their classifier. However, the code should execute if the user tries the drag event again. I believe this null reference error only occurs rarely if there are many hands trying to drag a galaxy quickly at the same time.

chelseatroy:
> If this code fails to run, a user would not be able to drag a galaxy into their classifier. However, the code should execute if the user tries the drag event again. I believe this null reference error only occurs rarely if there are many hands trying to drag a galaxy quickly at the same time.

Rock on. Sounds good.

PR #88, opened 2019-07-18
Add Garamond and Fix Scroll Bug

### LINKED ISSUE #85: Correctly Import Garamond
The NSF logo still doesn't have the correct font aside it, even through it is trying to access Garamond font. This might be an error in importing the font itself.

wgranger:
Describe your changes.
This PR fixes two small issues.

Garamond wasn't showing up on the table itself, although it appeared on my Surface and other machines. I think this is because the table itself can't find the font, so I've packed Garamond into the codebase.

Also, there was a small bug where the images on the examples panel would stay scrolled to the left, causing the image below.

image

To replicate this in the current app, select a galaxy example on the side panel of the classifier. Scroll the images all the way to the left and close the panel. When you reopen the classifier, the screenshot above should appear. This PR fixes that issue.

Fixes #85 .

# Review Checklist

- [x] Are tests passing?


GalaxyZooTouchTable/GalaxyZooTouchTable/GalaxyZooTouchTable.csproj
```diff
@@ -386,6 +386,7 @@
     <None Include="App.Release.config">
       <DependentUpon>App.config</DependentUpon>
     </None>
+    <Resource Include="Fonts\Garamond.ttf" />
```
chelseatroy: Looks good. Would it make sense for us to do a quick pass on the other fonts and make sure they all show up on the touch table as well? Or is the issue that the text doesn't show up at all (as opposed to showing up in a default font)?

wgranger: Other fonts (Karla and Poppins) are both showing up, mainly due to the fact their .ttf files are packed straight into the app. Garamond is the only font that I didn't include a .tff file along with the app. I did this because Garamond is a common font that should be on every system. However, it isn't showing up on the touch table itself, whereas it is showing up on other machines. In the case of the touch table, it currently falls back on a default font.

chelseatroy: Sounds like we now have all the fonts packaged up with the app, then. Works for me!

PR #93, opened 2019-07-22
Add Sentry for Error Reporting

wgranger:
Describe your changes.
This PR allows the app to use [Sentry ](https://docs.sentry.io/error-reporting/quickstart/?platform=csharp) for automated error reporting. If a bug crashes the app, an email will be sent with the relevant error message and stack trace. This requires the app to use a Sentry wrapper at the entry point. A `SENTRY_DSN` env variable is also required on the system, which is explained in the README. The DSN value can be retrieved through Sentry.

**PLEASE MERGE THIS LAST**: I would like to test error reporting with an updated version of the app to ensure the current floor deploy is reporting as expected, especially as the entry point has changed.

Fixes # .

# Review Checklist

- [x] Are tests passing?


wgranger:
This is now ready for review and has been tested and rebased with the current master.

wgranger:
It's set up now! Good thinking.

PR #101, opened 2019-08-06
Update logger

### LINKED ISSUE #100: Update Logs
The logs need to be tweaked a bit to include more information.

- Use a more human readable timestamp, perhaps using `Timestamp.Now` with millisecond precision in lieu of ticks.
-  Include subject ids when getting a random galaxy from `Random_Galaxy` and `Tap_Drop_Zone`
-  Galaxy titles aren't appearing in logs when choosing a galaxy example from the right pullout
-  Include which answer a user selected on submission
-  The Move_Map event might not be logging

I'll also include an article in the wiki entitled "Interpreting the Log Files" to further explain context behind each logged event.

wgranger:
Describe your changes.
This PR makes a few changes to the logger in order to include some new information we want to track and fix a couple bugs outlined in the associated issue.

This PR also changes the publishing certificate title as ours expired on August 2. I'll write up a bit of information on this in the [wiki](https://github.com/zooniverse/galaxy-zoo-touch-table/wiki/Troubleshooting:-Deployment-and-Updating) (see final section).

Fixes #100  .

# Review Checklist

- [x] Are tests passing?

### DIFF
--- a/GalaxyZooTouchTable/GalaxyZooTouchTable/Lib/Log.cs
+++ b/GalaxyZooTouchTable/GalaxyZooTouchTable/Lib/Log.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.IO;
 using System.Threading;
 
@@ -104,7 +105,7 @@ namespace GalaxyZooTouchTable.Lib
         {
             lock (gate)
             {
-                String entryString = PrintLog(entry, user, subjectId, state, context, peer, DateTime.Now.Ticks);
+                String entryString = PrintLog(entry, user, subjectId, state, context, peer, DateTime.Now.ToString("HH:mm:ss.ffff"));
                 activeEntryBuffer.Add(entryString);
                 if (EnableDebugOutput) System.Diagnostics.Debug.WriteLine(entryString);
             }
@@ -171,7 +172,7 @@ namespace GalaxyZooTouchTable.Lib
             }
         }
 
-        public string PrintLog(string entry, string user, string subjectId, ClassifierViewEnum state, string context, string peer, long time)
+        public string PrintLog(string entry, string user, string subjectId, ClassifierViewEnum state, string context, string peer, string time)
         {
             string stateString = null;
             if (state == ClassifierViewEnum.SubjectView)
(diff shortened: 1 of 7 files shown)


GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/SpaceViewModel.cs
```diff
@@ -150,6 +150,7 @@ private void OnShowError(string message)
 
         private void OnMoveViewWest(object obj)
         {
+            GlobalData.GetInstance().Logger?.AddEntry("Move_Map", context: "West");
```
chelseatroy: Good call. This makes me wonder: are there other places in the app that we might anticipate issues? Might be useful to have a list of "risky" spots to go back and add logs to if something starts happening that mystifies us. No change required to this PR, though.

GalaxyZooTouchTable/GalaxyZooTouchTable/Lib/Log.cs
```diff
@@ -93,18 +94,18 @@ public void AddHeaders()
...
             CheckActiveBuffer();
         }
 
-        public void AddEntry(string entry, string user = null, string subjectId = null, ClassifierViewEnum state = ClassifierViewEnum.None, string context = null, string peer = null)
+        public void AddEntry(string entry, string user = null, string subjectId = null, ClassifierViewEnum state = ClassifierViewEnum.None, string context = null, string peer = null, string answer = null)
         {
             lock (gate)
             {
-                String entryString = PrintLog(entry, user, subjectId, state, context, peer, DateTime.Now.Ticks);
+                String entryString = PrintLog(entry, user, subjectId, state, context, peer, answer, DateTime.Now.ToString("HH:mm:ss.ffff"));
```
chelseatroy: Times: the critical piece of information you don't need until you DO 😂

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/SpaceViewModel.cs
wgranger: There could be, and this comes up sometimes as we chat about the information we want to collect through the touch table. It's definitely something worth keeping in mind. The logs do a good job telling us what people are doing, but they also have the benefit of informing us that people are doing things we didn't anticipate (such as hitting the max level when classifying).

PR #102, opened 2019-08-08
Minor observations

### LINKED ISSUE #29: Leveling Max Out
Although it would be rare for a user to reach the max level (level five, 25 classifications) this will inevitably happen, and there should be some sort of max out state. We also discussed the possibility of something exciting happening whenever a user advances levels.

### LINKED ISSUE #94: Touchtable Observations: July 2019
Few items I noticed in recently observing the touch table:
- [ ] Timeouts should be shorter. Currently, the modal appears at 2:30. Perhaps it should be reduced by a minute?
- [ ] Moving the map should not cancel out all galaxies that are being dragged
- [ ] Some users might not be familiar with the terminology "drag and drop"

wgranger:
Describe your changes.
This PR makes a few minor changes based on observations we've seen recently on the table. I can't quite close #94 yet as it'll be a substantial change to keep galaxies from disappearing when the map moves.

**Changes here**:
- Reduce classifier timeout to two minutes. Still here modal opens at 1m 30s and classifier closes after two minutes of inactivity.
- Pulse center galaxies whenever a new user joins the table.
- Expand the hit area of the "Close Classifier" button.
- Allow leveler to left side of classifier to max out at "Master" level right past level five.
- Adjust a bug where the leveler wouldn't increment correctly.

Fixes #29  .
Working towards #94 

# Review Checklist

- [x] Are tests passing?

### DIFF
--- a/GalaxyZooTouchTable/GalaxyZooTouchTable/GalaxyZooTouchTable.csproj
+++ b/GalaxyZooTouchTable/GalaxyZooTouchTable/GalaxyZooTouchTable.csproj
@@ -53,10 +53,11 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <PropertyGroup>
-    <ManifestCertificateThumbprint>3F13CEEDA8A2CEE96FA4D0ECD0B5F57431DBABB9</ManifestCertificateThumbprint>
+    <ManifestCertificateThumbprint>D4A326B02C19A4F1BE50957DBF0F89E4A9E603EB</ManifestCertificateThumbprint>
   </PropertyGroup>
   <PropertyGroup>
-    <ManifestKeyFile>GalaxyZooTouchTable_TemporaryKey.pfx</ManifestKeyFile>
+    <ManifestKeyFile>
+    </ManifestKeyFile>
   </PropertyGroup>
   <PropertyGroup>
     <GenerateManifests>true</GenerateManifests>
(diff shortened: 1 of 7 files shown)


GalaxyZooTouchTable/GalaxyZooTouchTable/GalaxyZooTouchTable.csproj
```diff
@@ -388,9 +389,10 @@
     </None>
     <None Include="App.Release.config">
       <DependentUpon>App.config</DependentUpon>
+      <SubType>Designer</SubType>
     </None>
     <Resource Include="Fonts\Garamond.ttf" />
-    <None Include="GalaxyZooTouchTable_TemporaryKey.pfx" />
+    <None Include="GalaxyZooTouchTable_1_TemporaryKey.pfx" />
```
chelseatroy: This appears to change in a few different PRs. I assume we renamed something and that's propagating through the change sets?

GalaxyZooTouchTable/GalaxyZooTouchTable/ViewModels/LevelerViewModel.cs
```diff
@@ -97,24 +100,24 @@ public void Reset()
...
             LevelUpAnimation();
             IsOpen = true;
-            if (ClassificationLevel == MAX_LEVEL)
-            {
-                return;
-            }
             switch (ClassificationsThisSession)
             {
-                case int n when (n <= 6):
+                case int n when (n <= 5):
```
chelseatroy: I know that we set "One" to the default, but I wonder if there's a way to make that more explicit here (since it's the only level suspiciously missing from this piece of logic). There may not be.

GalaxyZooTouchTable/GalaxyZooTouchTable/GalaxyZooTouchTable.csproj
wgranger: Yes, I couldn't build and publish any of these changes without creating a new temporary key (that lasts for a year), so you'll see these changes in any of the current PRs until they hit master. There's more information about this in the repo's wiki under the troubleshooting section.

PR #110, opened 2019-11-27
Change Retirement Number

wgranger:
Describe your changes.
Just a couple small changes to set the retirement limit to 10. We've found out users generally do well on the touch table, so we can use a lower retirement limit to get through subject sets quicker.

Fixes # .

# Review Checklist

- [x] Are tests passing?


GalaxyZooTouchTable/GalaxyZooTouchTable/Models/ClassificationSummary.cs
```diff
@@ -17,6 +18,8 @@ public ClassificationSummary(string subjectLocation, ClassificationCounts counts
             SubjectLocation = subjectLocation;
             SummaryString = summaryString;
             TotalVotes = counts.Total;
+            if (TotalVotes > 10)
+                VoteLimit = counts.Total;
```
chelseatroy: Useful. I like that the VoteLimit takes into account the possibility that there are more votes than that limit.

PR #114, opened 2019-12-30
Don't Show Classify Screen if Subject Retired

### LINKED ISSUE #112: Retirement Modal Can Lead to Duplicate Classifications
The retirement modal sometimes shows over a previously shown subject, which may allow a user to classify a galaxy twice. This can occur through the following:

1. User classifies a galaxy and is looking at a classification summary.
2. Next, the user pulls an already classified galaxy onto the classifier.
3. Whoops, that subject is retired, so the classifier shows the "already retired" modal
4. In that process, the classification summary screen is reset back to the answer screen, but the subject is still the same as from step one (but with an answer selected).
5. At this point, a user could classify the same subject twice after closing the retirement modal.

wgranger:
Describe your changes.
Opening the retirement modal will not keep the classifier on the current state (showing answers or summary).

Fixes #112 

# Review Checklist

- [x] Are tests passing?


wgranger:
This is live on the floor with the correct version increment shown here.

